Issue #20, opened 2024-05-30
Use a sha in the HCL file instead of the main tag


tpendragon:
This is so we can roll back.

hackartisan:
closed by the `branch_or_sha` variable added in #17

PR #23, opened 2024-05-30
Fix sha

tpendragon:


hackartisan:
Does this reference #20? If so, how does it work; is there some documentation you could link to on this PR?

PR #24, opened 2024-06-01
Add deploy script.

### LINKED ISSUE #21: Write a script to deploy with nomad
It should get a token, notify slack, and update the tag to use the commit sha for the branch you ask it for (or use the sha you ask it for.)

It can be in ruby.

tpendragon:
Refs #21

Not sure this is the best script, and the output includes the SSH banner and stuff, but it's pretty simple!

This does Slack notifications by updating Github, which will notify slack since we've registered deployment notifications in #figgy!

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -7,6 +7,7 @@ then
   echo "Missing Environment. Command: BRANCH=main ./bin/deploy staging"
   exit
 fi
-ssh deploy@nomad-host-prod1.lib.princeton.edu << EOF
-  curl -s "https://raw.githubusercontent.com/pulibrary/${REPOSITORY}/${BRANCH_NAME}/config/deploy/${ENV}.hcl" | nomad job run -var "branch_or_sha=sha-$(git ls-remote https://github.com/pulibrary/${REPOSITORY}.git ${BRANCH_NAME} | awk '{ print substr($1,1,7) }')" -
+ssh -qtt deploy@nomad-host-prod1.lib.princeton.edu << EOF
+  clear
+  curl -s "https://raw.githubusercontent.com/pulibrary/${REPOSITORY}/${BRANCH_NAME}/config/deploy/${ENV}.hcl" | nomad job run -var "branch_or_sha=sha-$(git ls-remote https://github.com/pulibrary/${REPOSITORY}.git ${BRANCH_NAME} | awk '{ print substr($1,1,7) }')" - && logout
 EOF
--- a/bin/deploy
+++ b/bin/deploy
@@ -1,0 +1,12 @@
+#!/bin/bash
+ENV=$1
+BRANCH_NAME="${BRANCH:-main}"
+REPOSITORY="${REPO:-dpul-collections}"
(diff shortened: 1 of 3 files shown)


bin/deploy
```diff
@@ -0,0 +1,22 @@
...
+JOB_NAME="${JOBNAME:-dpulc}"
+if [[ -z ${ENV} ]];
+then
+  echo "Missing Environment. Command: BRANCH=main ./bin/deploy staging"
+  exit
+fi
+# Notify Github of Starting Deploy, Deploy, Notify Github of Ending Deploy
+ssh -qtt deploy@nomad-host-prod1.lib.princeton.edu << EOF
+  DEPLOY_ID=\$(curl -X POST -H "Accept: application/vnd.github+json-H" -H "Content-Type: application/x-www-form-urlencoded" -H "Authorization: Bearer \$GITHUB_TOKEN" --data "{\"ref\":\"${BRANCH_NAME}\",\"description\":\"Deploy from Nomad script\", \"auto_merge\": false, \"environment\": \"${ENV}\", \"required_contexts\": [] }" "https://api.github.com/repos/pulibrary/${REPOSITORY}/deployments" | jq .id)
+  curl -s -X POST -H "Accept: application/vnd.github+json-H" -H "Content-Type: application/x-www-form-urlencoded" -H "Authorization: Bearer \$GITHUB_TOKEN" --data "{\"environment\":\"${ENV}\",\"state\":\"in_progress\",\"log_url\": \"https://nomad.lib.princeton.edu/ui/jobs/${JOB_NAME}-${ENV}\", \"description\":\"Deployment finished successfully.\"}" "https://api.github.com/repos/pulibrary/${REPOSITORY}/deployments/\$DEPLOY_ID/statuses"
```
hackartisan, line 14: Can we add an empty line above? It sort of looks initially like a part of the line above.

hackartisan, line 14: typo "deployment finished successfully" should be "deployment started" or something.

bin/deploy
```diff
@@ -0,0 +1,22 @@
...
+BRANCH_NAME="${BRANCH:-main}"
+REPOSITORY="${REPO:-dpul-collections}"
+JOB_NAME="${JOBNAME:-dpulc}"
+if [[ -z ${ENV} ]];
+then
+  echo "Missing Environment. Command: BRANCH=main ./bin/deploy staging"
+  exit
+fi
+# Notify Github of Starting Deploy, Deploy, Notify Github of Ending Deploy
+ssh -qtt deploy@nomad-host-prod1.lib.princeton.edu << EOF
```
hackartisan, line 12: Add a comment: Run the github commands from the host machine because the pulbot token is set up there for the deploy user.

hackartisan (changes requested):
We're looking together at this and one thing that @tpendragon noted is that the deploy never fails. Trey's suggestion is that maybe we can get the exit status from the nomad job command, which runs synchronously to the whole job.

hackartisan (approved):
🎉

Issue #30, opened 2024-06-18
Develop an Indexing Pipeline


tpendragon:
This will have a bunch of sub-issues, this is mostly to record the fact that we need to make this.

## Acceptance Criteria

- [x] #41
- [x] #42
- [x] #43
- [x] #44
- [x] #45
    * Indexing phase should not kill Solr. Have some way to throttle traffic to Solr.
- [ ] #46
- [x] #47

## First Step

Split those tickets up up there.

Finalize the ADR: https://github.com/pulibrary/dpul-collections/pull/34

## Architecture Discussions

https://docs.google.com/document/d/1VgYeonai-q4NI6yXK0n3rYAipDsHH7j85C3XxoP-W6E/edit#heading=h.8n3o7dn0pcfs

## Content

All public Ephemera resources that are images

hackartisan:
close and let #46 stand on its own.

PR #34, opened 2024-07-09
Add first draft of indexing ADR.

tpendragon:
Refs #30

### DIFF
--- a/architecture-decisions/0002-indexing.md
+++ b/architecture-decisions/0002-indexing.md
@@ -79,15 +79,9 @@ Every minute Reindex will poll the records cached by the Transformation step and
 
 We expect reindexing to need to happen often - either because of changing weights in Solr, migrating Solr machines, or testing new configurations. By tightening up this time as much as possible we can try many different weights in a day, supporting our vision of being able to create a joyful discovery experience. We believe this performance estimate is reasonable given that there won't be any transformation necessary - it will go as fast as Solr can accept documents.
 
-## Consequences
+## Sequence Diagram
 
-We need to find a way to validate that we're indexing 100% of the documents that we pull from Figgy.
-
-Keeping track of three different tables may be complicated. However, we expect to be able to scale this architecture out to allow for multiple harvest sources and transformation steps in the future.
-
-
-## WIP full diagram
-
+```mermaid
 ---
 title: A full Indexing Pipeline workflow
 ---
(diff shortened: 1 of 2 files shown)


architecture-decisions/0002-indexing.md
```diff
@@ -0,0 +1,86 @@
...
+
+The faster we can do a full re-harvest, the faster we can pull in broad metadata changes from upstream (such as new Figgy or Bibdata data.) We want these kinds of tickets to have at most two days of delay.
+
+### Transformation
+
+Every minute Transformation will poll the records cached by the Hydration step, convert them to a Solr document, and store that solr document in a local postgres cache with the following structure:
+
+| id   | data  | log_order |
+|------|-------|-----------|
+| UUID | JSONB | INT       |
```
hackartisan, line 60: Is this the same log order value as the previous table, or different?

architecture-decisions/0002-indexing.md
```diff
@@ -0,0 +1,86 @@
...
+
+Every minute Reindex will poll the records cached by the Transformation step and index them into Solr as a batch.
+
+#### Performance Requirements
+
+10 minutes - 1 hour
+
+##### Performance Reasoning
+
+We expect reindexing to need to happen often - either because of changing weights in Solr, migrating Solr machines, or testing new configurations. By tightening up this time as much as possible we can try many different weights in a day, supporting our vision of being able to create a joyful discovery experience. We believe this performance estimate is reasonable given that there won't be any transformation necessary - it will go as fast as Solr can accept documents.
```
hackartisan, line 80: There's still more to think about in the indexing step. How do we ensure we're not crashing other applications by flooding the solr box/es? Do we need changes to the existing solr cloud setup? Is there any of this that should be part of this ADR? Or is it all a separate concern?

architecture-decisions/0002-indexing.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Context
+
+DPUL-Collections must have a resilient indexing pipeline that can quickly harvest, transform, and index millions of records. We foresee needing to index millions of records, regularly change weighting algorithms, and accept records from external institutions which may not be stable in the long term.
+
+There must be a verifiable method of ensuring that 100% of Figgy's relevant records are indexed into DPUL-Collections, to prevent us from constantly scrambling and diagnosing indexing issues as we do now with our spotlight-powered DPUL.
+
+We will be starting with indexing from Figgy, so that's where our initial performance requirements will be based upon.
+
+Often times systems like this use event streaming platforms such as Kafka, but we'd like to prevent adding new technology to our stack. We think we can use Postgres tables as a compact event log.
```
hackartisan, line 17: Perhaps add that we rely on figgy's single-node database architecture for the integrity of our event log

architecture-decisions/0002-indexing.md
```diff
@@ -0,0 +1,86 @@
...
+
```
hackartisan, line 18: Maybe reference the architecture book we read?

hackartisan (changes requested):
This is very clearly laid out, and I like the way the performance reasoning is in each section.

This descriptions in this document focus primarily on incremental changes (although I think the performance metrics are oriented around a full cache refresh in each case). We haven't talked as much about a full refresh of all 3 caches. Should we try to futher specify that process more here or address it separately?

tpendragon:
> This descriptions in this document focus primarily on incremental changes

That's interesting, it was totally my intent that this ADR be mostly about a full cache refresh, with a note that incremental changes should be much faster. I'll have to re-read with this viewpoint in mind.

architecture-decisions/0002-indexing.md
tpendragon: > There's still more to think about in the indexing step. How do we ensure we're not crashing other applications by flooding the solr box/es? Do we need changes to the existing solr cloud setup? Is there any of this that should be part of this ADR? Or is it all a separate concern?

:+1: Is "the indexing system should implement a way to apply back-pressure to allow us to scale the amount of processing we're demanding of Solr, to prevent our indexing from adversely affecting other applications" total nonsense or does it make sense?

hackartisan: it sounds good; I assume you have some idea of how to do that?

hackartisan:
> > This descriptions in this document focus primarily on incremental changes
> 
> That's interesting, it was totally my intent that this ADR be mostly about a full cache refresh, with a note that incremental changes should be much faster. I'll have to re-read with this viewpoint in mind.

I think it's that each description says "every minute..."

Also I think it would be good to talk about the life of the event log. if we do a full cache refresh, then another one, does the event log just get bigger? Do we know where to start to read the latest of everything?

architecture-decisions/0002-indexing.md
tpendragon: Uh, a few years ago it was "Flow": https://hexdocs.pm/flow/Flow.html. I dunno what the world is like now.

tpendragon: Or Broadway with a custom producer connected to ecto? https://github.com/dashbitco/broadway and something local like https://github.com/maartenvanvliet/off_broadway_ecto

tpendragon:
Oh good read, we need another column that's a consistent ID so we can compact it. So that if it exists already we UPDATE.

hackartisan (approved):

PR #59, opened 2024-07-24
43 hydration

tpendragon:


test/support/figgy_test_support.ex
```diff
@@ -0,0 +1,13 @@
+defmodule FiggyTestSupport do
```
hackartisan, line 1: Should this go in test/support/fixtures?

eliotjordan: The `markers` function is a fixture, but the other two functions are not. Moving the fixtures into a properly namespaced module is a good idea.

eliotjordan (approved):

PR #67, opened 2024-08-09
Don't cleanup figgy-fixtures.

tpendragon:


eliotjordan (approved):

hackartisan:
I think we wanted to exclude this because it's generated manually, not automated.

PR #72, opened 2024-08-21
Add console script.

tpendragon:
This'll open a bash shell or elixir console into a running container.


hackartisan (changes requested):
This is awesome!

suggestions:
- maybe use `iex` instead of `console` for the iex option? it's shorter and I think a bit more specific.
- put something in the readme about it!

tpendragon:
> * maybe use `iex` instead of `console` for the iex option? it's shorter and I think a bit more specific.

I thought about that, but I'll admit I'm thinking about what the world looks like when we're running rails containers too.

hackartisan (approved):

PR #77, opened 2024-08-22
Add an integration test that hydrates the full data set.

tpendragon:
This adds an integration test that does everything by kicking off all resources and waiting for the last one to show up in the database. It spawns a process with `Task.async` so we can set a timeout, so if it's too slow it crashes.

This test takes something like three seconds, so either we have to be okay with that for now, or find a way to be happy with unit tests for this filtering functionality somehow.


sdellis (approved):
LGTM

PR #90, opened 2024-09-18
Poll for updates

### LINKED ISSUE #91: Hydration, translation, and indexing should run simultaneously
The pipelines should all start at the same time and feed from one to the next. The original integration test ran each step consecutively one at a time.

This requires adding polling to the translation and indexing broadways.

tpendragon:
closes #91


test/dpul_collections/indexing_pipeline_test.exs
```diff
@@ -30,6 +30,38 @@ defmodule DpulCollections.IndexingPipelineTest do
...
+        IndexingPipeline.write_hydration_cache_entry(%{
+          data: %{},
+          source_cache_order: ~U[2024-07-22 20:05:00Z],
+          cache_version: 0,
+          record_id: "some record_id"
+        })
+
+      reloaded = IndexingPipeline.get_hydration_cache_entry!(record_2.id)
+      assert record.cache_order != reloaded.cache_order
+      record_3 = IndexingPipeline.get_hydration_cache_entry!(record_2.id)
```
hackartisan, line 61: Something doesn't look right in this test. we're fetching record_3 from record_2's id here? Also they all have the same `record_id` value -- I don't think that's relevant but I think it's a little confusing.

PR #101, opened 2024-09-25
Add functions for reindexing

### LINKED ISSUE #47: Create and document a full rehydration, retransformation, and reindexing process
## Acceptance Criteria

- [ ] There's a way to rehydrate (resulting in retransformation + reindexing) using the same cache_version
- [ ] There's a way to to retransform (resulting in reindexing) using the same cache_version
- [ ] There's a way to reindex using the same cache_version
- [ ] Documentation on how to do the above.

# First Step

Read the code implemented for #30

eliotjordan:
Adds functions for deleting processor marker and restarting the producer (clearing state) for each indexing stage.

closes #47


lib/dpul_collections/utilities.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  def reindex_hydration_cache(cache_version) do
+    hydration_processor_marker =
+      IndexingPipeline.get_processor_marker!("hydrator", cache_version)
+
+    IndexingPipeline.delete_processor_marker(hydration_processor_marker)
+    GenServer.stop(Figgy.HydrationConsumer)
+  end
+
+  def reindex_all(cache_version) do
```
tpendragon, line 31: Do we need to reset the state of everything if want to reindex? Like if we want a full re-hydration, can't we just reset the hydrator? And if we want a full retransformation without hydration, we just reset the transformer?

lib/dpul_collections/utilities.ex
```diff
@@ -0,0 +1,36 @@
...
+    # This resets the state and last queried marker.
+    GenServer.stop(Figgy.IndexingConsumer)
+  end
+
+  def reindex_transformation_cache(cache_version) do
+    transformation_processor_marker =
+      IndexingPipeline.get_processor_marker!("figgy_transformer", cache_version)
+
+    IndexingPipeline.delete_processor_marker(transformation_processor_marker)
+    GenServer.stop(Figgy.TransformationConsumer)
```
tpendragon, line 20: Should we use Broadway.stop?

lib/dpul_collections/utilities.ex
```diff
@@ -0,0 +1,36 @@
+defmodule DpulCollections.Utilities do
+  alias DpulCollections.IndexingPipeline
+  alias DpulCollections.IndexingPipeline.Figgy
+
+  def reindex_solr(cache_version) do
```
tpendragon, line 5: Should these functions be part of the Consumer namespace? So like, `Figgy.IndexingConsumer.Restart()` or `Reset()` or `ResetTo(nil)` or something?

tpendragon (dismissed):
More discussion points than specific changes.

tpendragon:
Update: I tried doing it this way while I was reindexing for description and it didn't work - if indexing is running, it writes the marker between me updating and restarting the node, no matter how quickly I went. I think we need to stop the producer in-band somehow.

hackartisan:
> Update: I tried doing it this way while I was reindexing for description and it didn't work - if indexing is running, it writes the marker between me updating and restarting the node, no matter how quickly I went. I think we need to stop the producer in-band somehow.

I feel like we should be able to start these broadway pipelines as named processes so that we can shut them down by referencing them with an atom instead of needing a PID

test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
```diff
@@ -1,10 +1,10 @@
 defmodule DpulCollections.IndexingPipeline.FiggyFullIntegrationTest do
+  alias ElixirLS.LanguageServer.Providers.CodeLens.Test
```
hackartisan, line 2: Remove this?

hackartisan (changes requested):

hackartisan (dismissed):

eliotjordan (commented):
Nice work. 💯

PR #106, opened 2024-09-27
Add info about where to find logs

hackartisan:


eliotjordan (approved):

tpendragon (approved):
Looks good. I wonder if we want a `/bin/logs staging tail` or something.

hackartisan:
Yes I think we'll want log aggregation. I'll make a ticket for that.

PR #109, opened 2024-09-27
Combine TransformationCacheEntryMarker & HydrationCacheEntryMarker

tpendragon:
They're both the same concept, and all the code was exactly the same.


tpendragon:
Closing because I like #110 much more.

PR #110, opened 2024-09-27
Proof of concept for combining Transformer/Indexing/Hydration Producers

tpendragon:
They both do the same thing and the code was all the same except two
methods. This is one way I was able to combine them.


tpendragon:
Yeah I agree - a documented behavior for those modules to adhere to would be great

tpendragon:
@hackartisan Okay. Now this does a couple extra things:

1. It adds the concept of a `Source`, which is the behaviour that the IndexingProducerSource, HydrationProducerSource, and TransformationProducerSource
2. It gets rid of the HydrationProducer & RecordCacheEntry (turns out it was also just all duplicated)

hackartisan (approved):
I like this very much. I'm holding off on merging to give other people on the team a chance to weigh in.

PR #116, opened 2024-10-03
Descriptions w/ Language Detection

### LINKED ISSUE #111: Index description for Ephemera Folders
## Acceptance Criteria

- [ ] There's a description field in the solr document which is indexed and searchable.

## First Step

- [ ] Figure out which field in Ephemera maps to "description" and find an example record.

tpendragon:
This uses Solr's language detection features to generate a description field for every language it could be, so we can hopefully improve searching for that field, given that we know it's going to be in multiple languages.

I do not know how we'll actually enable search for all these languages, but at least this would get us them separated out.

Solr language detection: https://solr.apache.org/guide/7_2/detecting-languages-during-indexing.html

Closes #111

### DIFF
--- a/test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
+++ b/test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
@@ -87,7 +87,7 @@ defmodule DpulCollections.IndexingPipeline.FiggyFullIntegrationTest do
     # Language detection
     assert %{"description_txtm_en" => [first_description | _tail]} = document
     assert first_description |> String.starts_with?("Asra-Panahi") == true
-    assert %{"language_ss" => ["en"]} = document
+    assert %{"detectlang_ss" => ["en"]} = document
 
     hydrator |> Broadway.stop(:normal)
     transformer |> Broadway.stop(:normal)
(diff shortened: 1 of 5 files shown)


test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
```diff
@@ -71,4 +84,62 @@ defmodule DpulCollections.IndexingPipeline.FiggyFullIntegrationTest do
...
+  test "indexes description" do
+    {hydrator, transformer, indexer, document} =
+      index_record_id("26713a31-d615-49fd-adfc-93770b4f66b3")
+
+    assert %{"description_txtm" => [first_description | _tail]} = document
+    assert first_description |> String.starts_with?("Asra-Panahi") == true
+    # Language detection
+    assert %{"description_txtm_en" => [first_description | _tail]} = document
+    assert first_description |> String.starts_with?("Asra-Panahi") == true
+    assert %{"language_ss" => ["en"]} = document
```
hackartisan, line 139: Is this language field specific to the description?

tpendragon: I think if we detect languages in more fields it'll append that here

hackartisan (commented):
Was there another resource you used in setting this up? The docs linked explain the specific lang detection settings but didn't mention setting up all these fields, the regex jar file, or the update chain configuration.

tpendragon:
> Was there another resource you used in setting this up? The docs linked explain the specific lang detection settings but didn't mention setting up all these fields, the regex jar file, or the update chain configuration.

That's because the documentation is super light! I had to figure out the jar file myself, seems undocumented. Without it, Solr errored about not being able to find a class. The update chain config is a combo of that doc that's shown and https://solr.apache.org/guide/8_2/update-request-processors.html.

hackartisan (approved):

PR #117, opened 2024-10-04
Cluster DPUL-Collections nodes w/ DNS

### LINKED ISSUE #118: DPUL-Collections machines should network with one another
Connecting the machines via Elixir networking will let Phoenix.PubSub, Phoenix.Presence, and LiveView be aware of multiple machines and share information automatically. It'll also let us write distributed software in the future.

## Acceptance Criteria

- [ ] `Node.list()` on one DPUL-C node returns the address of the other nodes that are launched.

## First Step

So much research.

## Technical Tips

(This was written after the fact, I'm mostly writing it here to include some info)
1. You can generate setup files for releases via `mix release.init`
1. Releases need a name, defined by `RELEASE_NODE`.
1. RELEASE_DISTRIBUTION needs to be `name`
1. https://github.com/phoenixframework/dns_cluster expects nodes to be named `<app name>@<machine ip>`
1. Erlang usually uses EPMD (Erlang Port Mapper Daemon) for networking, which is a daemon that runs on port 4369 and records a random high-number port for each erlang VM on the machine that other erlang VMs should use for distributed traffic.
1. You can turn off EPMD and instead just set the port used for distributed traffic by setting `-start_epmd false -erl_epmd_port 6789`.
1. We can rely on consul for DNS discovery of new nodes, but to do so containers need to be able to run consul queries. We can do that by setting the container's DNS to be the machine it's running on (which has a running consul DNS interface.)

tpendragon:
This does the following:

1. Disables EPMD (Erlang Port Mapper Daemon)
2. Makes all distribution traffic go through a known port (6789).
3. Forwards that port for each web instance.
4. Uses the DNS server on the host so it can run consul queries.
5. Uses a consul DNS query to get all hosts running DPUL-C.
6. Configures dns_cluster to query those hosts.

Companion PR: https://github.com/pulibrary/princeton_ansible/pull/5424

Closes #118

### DIFF
--- a/config/deploy/staging.hcl
+++ b/config/deploy/staging.hcl
@@ -40,6 +40,11 @@ job "dpulc-staging" {
         timeout = "1s"
       }
     }
+    affinity {
+      attribute = "${meta.node_type}"
+      value = "default"
+      weight = 100
+    }
     task "release-migrate" {
       # The dbmigrate task will run BEFORE the puma task in this group.
       lifecycle {
@@ -109,6 +114,11 @@ job "dpulc-staging" {
     network {
       port "http" { to = 4000 }
       port "epmd" { static = 6789 }
+    }
+    affinity {
+      attribute = "${meta.node_type}"
+      value = "worker"
+      weight = 100
     }
     service {
       name = "dpulc-staging-web"
--- a/config/deploy/staging.hcl
+++ b/config/deploy/staging.hcl
@@ -65,6 +65,7 @@ job "dpulc-staging" {
         SECRET_KEY_BASE = {{ .SECRET_KEY_BASE }}
         CACHE_VERSION = ${var.cache_version}
         PHX_HOST = ${var.host}
+        DNS_CLUSTER_QUERY = "dpulc-staging-web.service.consul"
         {{- end -}}
         EOF
       }
@@ -107,6 +108,7 @@ job "dpulc-staging" {
     count = 1
     network {
       port "http" { to = 4000 }
+      port "epmd" { static = 6789 }
     }
     service {
       name = "dpulc-staging-web"
--- a/config/deploy/staging.hcl
+++ b/config/deploy/staging.hcl
@@ -126,6 +126,10 @@ job "dpulc-staging" {
         ports = ["http"]
         force_pull = true
       }
+      env {
+        RELEASE_IP = "${NOMAD_IP_http}"
+        ERL_DIST_PORT = 6789
+      }
       # Save a bunch of CPU and RAM to run indexing.
       resources {
         cores = 6
--- a/config/deploy/staging.hcl
+++ b/config/deploy/staging.hcl
@@ -136,7 +136,7 @@ job "dpulc-staging" {
       driver = "podman"
       config {
         image = "ghcr.io/pulibrary/dpul-collections:${ var.branch_or_sha }"
-        ports = ["http"]
+        ports = ["http", "epmd"]
         force_pull = true
       }
       env {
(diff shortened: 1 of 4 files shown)


hackartisan:
@tpendragon can you link this to an issue?

config/deploy/staging.hcl
```diff
@@ -61,9 +66,18 @@ job "dpulc-staging" {
       driver = "podman"
       config {
         image = "ghcr.io/pulibrary/dpul-collections:${ var.branch_or_sha }"
-        ports = ["http"]
+        ports = ["http", "epmd"]
         force_pull = true
       }
+      resources {
+        cpu    = 2000
+        memory = 1000
```
hackartisan, line 74: how did we chose these numbers? what is the unit? can we document a way to assess these numbers moving forward?

rel/env.sh.eex
```diff
@@ -0,0 +1,22 @@
...
+
+# # Set the release to load code on demand (interactive) instead of preloading (embedded).
+# export RELEASE_MODE=interactive
+
+# # Set the release to work across nodes.
+# # RELEASE_DISTRIBUTION must be "sname" (local), "name" (distributed) or "none".
+export RELEASE_DISTRIBUTION=name
+export RELEASE_NODE="<%= @release.name %>@${RELEASE_IP}"
+export ERL_DIST_PORT=6789
+# export RELEASE_NODE="<%= @release.name %>@127.0.0.1"
```
hackartisan, line 22: remove this commented code

lib/dpul_collections/indexing_pipeline/figgy/hydration_consumer.ex
```diff
@@ -4,6 +4,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationConsumer do
   """
   alias DpulCollections.IndexingPipeline
   alias DpulCollections.IndexingPipeline.Figgy
+  alias DpulCollections.IndexingPipeline.DatabaseProducer
```
hackartisan, line 7: might need to rebase

config/deploy/staging.hcl
```diff
@@ -15,9 +15,14 @@ job "dpulc-staging" {
...
   }
   group "web" {
-    count = 1
+    count = 2
     network {
       port "http" { to = 4000 }
+      port "epmd" { static = 6789 }
+      # Add the consul DNS loopback, so we can use consul queries.
+      dns {
+        servers = ["10.88.0.1", "128.112.129.209", "8.8.8.8", "8.8.4.4"]
```
hackartisan, line 24: Where did we get these?

hackartisan (changes requested):
still trying to understand details here.

see also question on the companion PR: https://github.com/pulibrary/princeton_ansible/pull/5424/files#r1801347657

config/deploy/staging.hcl
tpendragon: While indexing in Nomad CPU would never break 2ghz, memory wouldn't break 1 GB, so I adjusted them here. Nomad was killing the indexer for going above the previous defaults (which are like 500 mhz/300mb)

config/deploy/staging.hcl
tpendragon: The first one is the loopback address for podman, the second was DNS from the host, which I assume is the org's DNS server - I'm not sure if it's important now 🤔

config/deploy/staging.hcl
hackartisan: Maybe an ADR about this??

tpendragon: This got in main as part of the indexer, which has its own ADR. Good?

hackartisan (approved):

tpendragon:
This is broken.

hackartisan (approved):

PR #136, opened 2024-10-16
Add home page LiveView

### LINKED ISSUE #125: Implement home page slice wireframe
## Acceptance Criteria

- [x] The root URL looks roughly like the wireframe below.
- [ ] ~~the header / footer are included via the app layout rather than the root layout~~

## First Step

Create a route and live view for the home page.

## Wireframe
375095301-687a564f-e591-4381-a134-82cd7d3bbeb8

eliotjordan:
Work towards #125 

Screenshot 2024-10-17 at 2 11 12 PM


lib/dpul_collections_web/router.ex
```diff
@@ -13,7 +13,7 @@ defmodule DpulCollectionsWeb.Router do
   scope "/", DpulCollectionsWeb do
     pipe_through :browser
 
-    get "/", PageController, :home
```
hackartisan, line 16: you mentioned deleting the controller, etc, also. I would be in favor of that.

eliotjordan: Removed.

sdellis (approved):
LGTM

PR #138, opened 2024-10-16
Index dates

### LINKED ISSUE #134: Index dates for Ephemera Folders
## Acceptance Criteria

- [ ] There's a years field in the solr document that has all of the years in an array that should match.
- [ ] There's a display_date field in the solr document that is what will render in the record page.

## Implementation Note

The suffix isn't defined above, part of the ticket is figuring out which suffixes are appropriate for each field.

suffixes to use:
- display field should be string, `display_date_ss`
- years field, used for sorting, searching, and faceting, should be integer, `years_is`.

hackartisan:
closes #134 

I made the display date single valued. Open to discussion here.

- Refactor transformation logic into entry module
- Small corrections to comments
- Add years_is to index
- Add display_date_ss
- Mix format
- Add assertions for date solr fields to integration test
- Make display date single value

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
@@ -23,11 +23,31 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
     %{record_id: id} = hydration_cache_entry
     %{data: %{"metadata" => metadata = %{"title" => title}}} = hydration_cache_entry
     description = get_in(metadata, ["description"])
+    years = extract_years(metadata)
 
     %{
       id: id,
       title_ss: title,
-      description_txtm: description
+      description_txtm: description,
+      years_is: years
     }
   end
+
+  defp extract_years(%{"date_range" => [%{"start" => [start_year], "end" => [end_year]}]}) do
+    Enum.to_list(String.to_integer(start_year)..String.to_integer(end_year))
+  end
+
+  defp extract_years(%{"date_created" => []}) do
+    nil
+  end
+
+  defp extract_years(%{"date_created" => date}) do
+    Enum.map(date, &String.to_integer/1)
+  end
+
+  defp extract_years(%{}) do
+    # there's no date_created value
+    nil
+  end
+
 end
--- a/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
@@ -31,7 +31,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
       title_ss: title,
       description_txtm: description,
       years_is: years,
-      display_date_ss: display_date
+      display_date_s: display_date
     }
   end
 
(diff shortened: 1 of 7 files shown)


lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
```diff
@@ -17,4 +17,61 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
...
+  defp extract_years(%{"date_range" => [%{"start" => [start_year], "end" => [end_year]}]}) do
+    Enum.to_list(String.to_integer(start_year)..String.to_integer(end_year))
+  end
+
+  defp extract_years(%{"date_created" => []}) do
+    nil
+  end
+
+  defp extract_years(%{"date_created" => date}) do
+    Enum.map(date, &String.to_integer/1)
```
tpendragon, line 47: Is this...there are multiple dates sometimes? Should this be `dates`?

tpendragon (changes requested):
I like this a lot. There's just a couple nit-picky things that I think might help us in the future, but open to discussion on them.

lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
```diff
@@ -17,4 +17,61 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
     |> cast(attrs, [:data, :cache_version, :record_id, :source_cache_order])
     |> validate_required([:data, :cache_version, :record_id, :source_cache_order])
   end
+
+  @spec to_solr_document(%__MODULE__{}) :: %{}
+  def to_solr_document(hydration_cache_entry) do
+    %{record_id: id} = hydration_cache_entry
+    %{data: %{"metadata" => metadata = %{"title" => title}}} = hydration_cache_entry
+    description = get_in(metadata, ["description"])
+    years = extract_years(metadata)
```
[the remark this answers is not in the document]
hackartisan: I debated that too. Happy to try it that way.

lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
hackartisan: Yes it should :)

hackartisan: actually no I think it shouldn't. It's a single-value field in the figgy form.

hackartisan: Okay I expect it will be a single date. But it's stored in an array because figgy does that. And it's going into a multi-valued field. Solr will map it to multi-value even if we give it a single value, but is that easier to understand than just outputting it as an array here to begin with? I'm trying a few things and none of them seem super clear. I may just comment.

tpendragon: What about `"date_created" => [date]`?

hackartisan: 😂 gmta, see if you like it.

tpendragon: I do like it! Part of me wonders if `"date_created" => [date | _tail]` or something would more explicitly say "even if there's two dates somehow, that's a mistake and we don't care", but I think this is good.

tpendragon (dismissed):

lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
hackartisan: actually that's a good point. I was gonna say we'd get a match error in that case but, but currently if that did happen it would fall through to the next clause and return nil and we'd never know we lost that date.

tpendragon (approved):

PR #150, opened 2024-10-21
Index page count.

### LINKED ISSUE #121: Index number of pages in a resource
Our wireframes have search results and record pages that show multiple pages of the resource - to see what it looks like to have multiple we should index how many there are, even though we're using filler images for now.

## Acceptance Criteria

- [ ] There's a field `page_count_??` which has the number of pages in the resource.

## Technical Implementation

We should be able to extract this from the length of `member_ids` - at least for Ephemera.

## First Step

Write a test.

tpendragon:
Closes #121


test/support/fixtures/figgy_test_fixtures.ex
```diff
@@ -58,8 +87,6 @@ defmodule FiggyTestFixtures do
         }
       })
 
-    :timer.sleep(1)
```
hackartisan, line 61: We don't need these anymore?

hackartisan (changes requested):

test/support/fixtures/figgy_test_fixtures.ex
tpendragon: I don't think so. It got added in https://github.com/pulibrary/dpul-collections/pull/90 while I was trying to figure out that race condition that turned out to be because we were querying the wrong field.

hackartisan (approved):

PR #151, opened 2024-10-21
Title language detection & Tokenized Search

### LINKED ISSUE #142: Make title search multilingual
We need to do language detection on titles, and also we want to be able to match non-exact titles in english, e.g. we need a title field that does stemming.

Look to https://github.com/pulibrary/dpul-collections/pull/116 for reference

tpendragon:
Notes:

1. Language detection doesn't work on copied fields, so we need to index title_txtm, then copy to title_ss.
2. Language detection on short strings is a little wonky, but that's probably fine. For instance, it really thinks 'Ali Bagheri' is German.

Closes #142

### DIFF
--- a/test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
+++ b/test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
@@ -1,4 +1,5 @@
 defmodule DpulCollections.IndexingPipeline.FiggyFullIntegrationTest do
+  alias Phoenix.LiveDashboard.TitleBarComponent
   use DpulCollections.DataCase
 
   alias DpulCollections.Repo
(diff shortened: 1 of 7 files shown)


test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
```diff
@@ -1,4 +1,5 @@
 defmodule DpulCollections.IndexingPipeline.FiggyFullIntegrationTest do
+  alias Phoenix.LiveDashboard.TitleBarComponent
```
hackartisan, line 2: what does this do?

tpendragon: Where does this stuff keep coming from

hackartisan (approved):

eliotjordan (approved):

PR #171, opened 2024-10-24
[DRAFT] Logging infrastructure Nomad files

tpendragon:
I'm largely just testing these. It's a combination of https://atodorov.me/2021/07/09/logging-on-nomad-and-log-aggregation-with-loki/ and https://dev.to/justinepdevasia/docker-log-observability-analyzing-container-logs-in-hashicorp-nomad-with-vector-loki-and-grafana-4dp4.

Big gaps:

1. Grafana's weird, we'd have to get used to it.
2. Data persistence for Grafana/Loki
3. URL for Grafana
4. Auth decision

### DIFF
--- a/config/deploy/grafana.hcl
+++ b/config/deploy/grafana.hcl
@@ -1,0 +1,52 @@
+job "grafana" {
+  datacenters = ["dc1"]
+  type        = "service"
+  node_pool   = "staging"
+
+  group "grafana" {
+    count = 1
+
+    network {
+      port "grafana" {
+        to = 3000
+      }
+      dns {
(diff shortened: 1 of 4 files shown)


config/deploy/grafana.hcl
```diff
@@ -0,0 +1,84 @@
...
+
+  group "grafana" {
+    count = 1
+
+    network {
+      port "grafana" {
+        to = 3000
+      }
+      dns {
+        servers = ["10.88.0.1", "128.112.129.209", "8.8.8.8", "8.8.4.4"]
```
eliotjordan, line 14: question: How does grafana use the google dns servers? Is this config intentional?

tpendragon: It's intentional, but just a fallback. Probably not required.

Issue #104, opened 2024-09-26
Swap in a new full index


hackartisan:
Sometimes we will need to generate a new index from empty.

Scenario 1: We discover a bug that has caused us to index records that shouldn't be publicly viewable. The records themselves haven't changed, but when we fix the bug we need to create a new clean index.

Scenario 2: We need to migrate to a new major version of solr, which requires us to create a new clean index.

We'd like to be able to accommodate these scenarios without downtime. Solr provides some tooling for this (e.g. the solr api, aliases), or we can add configuration to local application.

## Acceptance criteria
- [x] There's a documented way to create a new collection and initiate indexing to it while the old collection continues to index from our pipeline and serve queries.
- [x] There's a documented way to figure out when it's time to swap indexes. It's okay if this involves some sample queries to run on the solr box.
- [x] There's a task we can run to swap in the new collection. If possible, automate this.
- [x] There's a task we can run to delete the old collection.

## First Step

Look at the scripts in pdc_describe - they do a process like this.
  * https://github.com/pulibrary/pdc_discovery/blob/fc7c3a5a9e2032dc28d09c29bfaf8292548f061c/lib/tasks/index.rake
  * https://github.com/pulibrary/pdc_discovery/blob/main/app/lib/indexing/solr_cloud_helper.rb#L3

Also play around with it.

tpendragon:
Close this ticket when we try it in staging.

PR #178, opened 2024-10-25
Use solr cloud in local tests, add additional collection and alias

hackartisan:
- **Update lando to use solr cloud**
- **Replicate new solr test setup in ci**

advances #104 

Implications of this PR are that in development and test solr instances, there is now basic auth in place. I added a credential "user:pass" to make logging in slightly simpler than it is in the container's default settings.


eliotjordan (approved):
💯

PR #182, opened 2024-10-28
Fixes CSS for Homescreen and Search Results

sdellis:
WIP: fixes responsive buttons and forms

### DIFF
--- a/lib/dpul_collections_web/live/search_live.ex
+++ b/lib/dpul_collections_web/live/search_live.ex
@@ -77,7 +77,7 @@ defmodule DpulCollectionsWeb.SearchLive do
       <div id="filters" class="grid md:grid-cols-[auto_300px] gap-2">
         <form id="date-filter" class="grid md:grid-cols-[150px_200px_200px] gap-2">
           <label class="col-span-1 self-center font-bold uppercase" for="sort-by">
-            filter by date: 
+            filter by date:
           </label>
           <input
             class="col-span-1"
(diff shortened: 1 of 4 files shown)


sdellis:
Screenshot 2024-10-31 at 2 46 45 PM
Screenshot 2024-10-31 at 2 46 31 PM
Screenshot 2024-10-31 at 2 46 07 PM
Screenshot 2024-10-31 at 2 45 51 PM

lib/dpul_collections_web/live/search_live.ex
```diff
@@ -133,8 +135,31 @@ defmodule DpulCollectionsWeb.SearchLive do
...
+        <img
+          class="h-[350px] w-[350px] md:h-[225px] md:w-[225px]"
+          src="https://picsum.photos/350/350/?random"
+        />
+        <img
+          class="h-[350px] w-[350px] md:h-[225px] md:w-[225px]"
+          src="https://picsum.photos/350/350/?random"
+        />
+      </div>
+      <h2 class="underline text-xl font-bold pt-4"><%= @item.title %></h2>
```
hackartisan, line 162: Seeing this as an h2 makes me curious; I would expect h1 to be the name of the site and h2 to be the title of the page. which if we had that would make these h3. We don't have a page title in our design, but would it be good to have one hidden from screen for accessibility reasons?

hackartisan (changes requested):
This looks really good, I just had one thought!

lib/dpul_collections_web/live/search_live.ex
sdellis: @hackartisan that's a good point. I don't think we need to use the <h1> on every page for the name of the site, but the title of the page is important. In this case, I will add an <h1 class="sr_only"> to the page. Each result can then stay as an <h2>.

sdellis:
Added Page Count feature depicted in wireframes.

Screenshot 2024-10-31 at 3 58 59 PM
Screenshot 2024-10-31 at 3 58 17 PM
Screenshot 2024-10-31 at 3 57 33 PM

tpendragon (approved):

hackartisan (approved):
Yay!

Issue #124, opened 2024-10-09
Implement show page wireframe


tpendragon:
## Acceptance Criteria

- [ ] Going to the url for a record (`/item/:id`) results in a view that looks like below.

## First Step

Create a route and live view for a record page. 

## Wireframe Notes

The big box should be a big preview image generated by a random image generator - we recommend https://random-d.uk/api/randomimg, but go with your heart.

Every little box should also be a random image. Again, ducks are cute.

The download link should link to a download of the image. In this case...I guess another random duck picture.

## Wireframes

Full screen view:
![dpul-c-record-wireframe4](https://github.com/user-attachments/assets/f53a1ba1-46d6-4f9b-9182-1156ec26b14d)

Mobile view:
![dpul-c-record-page-mobile2](https://github.com/user-attachments/assets/b8d029f6-e5b5-45d6-8ff1-850c5657594f)

sdellis:
Updated to include `Download` and `View` buttons. Since individual pages are downloadable, this seems logical. View would open the item in a viewer.

![dpul-c-record-wireframe3](https://github.com/user-attachments/assets/36fb1390-cdd4-4773-82ce-1827d41b9434)

sdellis:
Mobile view:
![dpul-c-record-page-mobile-wireframe](https://github.com/user-attachments/assets/4a7d0061-7834-40cb-a84e-8b91306235b8)

sdellis:
Final wireframe for 1st "slice" Record Page.

Full screen view:
![dpul-c-record-wireframe4](https://github.com/user-attachments/assets/f53a1ba1-46d6-4f9b-9182-1156ec26b14d)

Mobile view:
![dpul-c-record-page-mobile2](https://github.com/user-attachments/assets/b8d029f6-e5b5-45d6-8ff1-850c5657594f)

sdellis:
Implements #29

hackartisan:
Still needed: Description, larger font for title and in general for heading fonts

PR #195, opened 2024-10-31
Add initial item show page and item routing

eliotjordan:
Work towards #124


lib/dpul_collections/solr.ex
```diff
@@ -10,11 +10,22 @@ defmodule DpulCollections.Solr do
     response.body["response"]["numFound"]
   end
 
+  @query_field_list [
```
hackartisan, line 13: we might want to push this to the solr config at some point; it will probably get long

hackartisan (approved):
This looks great.

test/dpul_collections_web/live/item_live_test.exs
```diff
@@ -0,0 +1,82 @@
...
+    {:ok, view, _html} = live(conn, "/item/2")
+    response = render(view)
+    assert response =~ "زلزلہ"
+    assert response =~ "2024"
+    assert response =~ "14"
+  end
+
+  test "/i/{:slug}/item/{:id} does not redirect with a bad id", %{conn: conn} do
+    conn = get(conn, "/i/not-a-real-slug/item/badid1")
+    response = html_response(conn, 200)
```
sdellis, line 73: @eliotjordan I'm curious about the 200 response for items that are not found. It's not a 404?

PR #204, opened 2024-11-01
Hydrator should transform in handle_message, save in handle_batch

### LINKED ISSUE #183: Consumers should transform in `handle_message` and persist in `handle_batch`
## Acceptance Criteria

- [ ] A message is transformed in `handle_message`, but saved in `handle_batch`

## Reasoning

As per https://hexdocs.pm/broadway/Broadway.html#c:handle_batch/4, when one message fails in `handle_batch` EVERY message is marked as failed. Since some items can't convert dates, every record in that batch of 50 is failing, resulting in drastically different counts than we expected.

## First Step

Brainstorm some ways to do this.

## Technical Info

One limitation here is that we expect to be able to acknowledge the message that we sent. It might be easiest if we started producing and acknowledging markers, then expecting consumers to resolve those markers into resources.

tpendragon:
Closes #183


hackartisan (approved):
🎉

PR #215, opened 2024-11-06
Adds page and component transitions

sdellis:
- Old Page fade-out before load 
- New Page fade-in when mounted
- WIP: Search results pagination transitions

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -3,12 +3,23 @@
 @import "tailwindcss/utilities";
 
 /* This file is for your main application CSS */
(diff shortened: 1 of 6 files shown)


tpendragon (changes requested):
I...think I'm okay with the image transitions, but when I deployed this to staging it made paginating feel slow. not because it actually was, but because loading the transition actually took longer than it takes for the page to actually load.

tpendragon:
We decided this was a good learning experience, but we're gonna add transitions when there's long loading.

PR #222, opened 2024-11-07
POC: Render indexed images.

tpendragon:
This is just a proof of concept to prove to myself indexing like this will work.

As an aside, I'm INSPIRED. I found myself adding features to this just to see the images.


lib/dpul_collections/item.ex
```diff
@@ -23,7 +23,7 @@ defmodule DpulCollections.Item do
       date: doc["display_date_s"],
       page_count: doc["page_count_i"],
       url: generate_url(id, slug),
-      image_service_urls: doc["image_service_urls_ss"]
+      image_service_urls: doc["image_service_urls_ss"] || []
```
hackartisan, line 26: Do we have a test for this case?

hackartisan (changes requested):
Okay so this renders the first page both as the large image and also below, but that makes sense because we'll slot the designated thumbnail image into the large image and that may not always be the same as the first page.

My only question really is for the last 2 commits here, what happened that showed we needed those and do we need regression test coverage for them?

hackartisan (changes requested):
This looks great! Thank you for the new tests. What is the "approximate" scenario?

tpendragon:
> This looks great! Thank you for the new tests. What is the "approximate" scenario?

Copy/paste. :) Fixed.

hackartisan (approved):

PR #225, opened 2024-11-08
Updated show page.

### LINKED ISSUE #124: Implement show page wireframe

tpendragon:
* Adds description
  - We said we'd add ellipses, but the CSS for this turned out to be pretty impossible - CSS ellipses is mostly for single-lines, and requires an absolute width.
  - This makes two divs for description so that in mobile it's below the image, but in desktop it's below the title. It's not the cleanest thing, but it works pretty effectively.
* Increases header sizes for the show page.

Closes #124

Increased headers:
Screen Shot 2024-11-08 at 1 00 06 PM

Mobile description after thumbnail:
Screen Shot 2024-11-08 at 1 00 21 PM

Long description:
Screen Shot 2024-11-08 at 12 59 51 PM

### DIFF
--- a/lib/dpul_collections/item.ex
+++ b/lib/dpul_collections/item.ex
@@ -7,7 +7,8 @@ defmodule DpulCollections.Item do
     :date,
     :page_count,
     :url,
-    :image_service_urls
+    :image_service_urls,
+    :description
   ]
 
   def from_solr(nil), do: nil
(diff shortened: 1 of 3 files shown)


hackartisan (approved):
The double description thing is confusing and we don't like it but it works. maybe someone will want to mess with refactoring it later.

PR #235, opened 2024-11-11
[Draft] Fun experiment - add some random selections to homepage.

tpendragon:
Just curious what this looks like.


lib/dpul_collections_web/live/search_live.ex
```diff
@@ -126,7 +124,13 @@ defmodule DpulCollectionsWeb.SearchLive do
     """
   end
 
-  attr :item, Item, required: true
+  def items(assigns) do
```
hackartisan, line 127: this seems like a nice refactor, though!

PR #240, opened 2024-11-13
WIP: adds localization features

sdellis:
still working out how to pass the locale parameter to every liveview page


lib/dpul_collections_web/components/lux-components.ex
```diff
@@ -1,16 +1,26 @@
...
+      app-url="/"
+      theme="shade"
+    >
+      <lux-menu-bar
+        type="main-menu"
+        v-bind:menu-items={"[
+        {name: '#{gettext("Language")}', component: 'Language', children: [
           {name: 'English', component: 'English', href: '?locale=en'},
           {name: 'Español', component: 'Español', href: '?locale=es'},
           {name: 'Português do Brasil', component: 'Português do Brasil', href: '?locale=pt-BR'}
```
hackartisan, line 19: Should we take out the Brazilian until we get the translations in place?

lib/dpul_collections_web/live/home_live.ex
```diff
@@ -1,8 +1,11 @@
 defmodule DpulCollectionsWeb.HomeLive do
   use DpulCollectionsWeb, :live_view
+  import DpulCollectionsWeb.Gettext
   alias DpulCollections.Solr
   alias DpulCollectionsWeb.Live.Helpers
 
+  # on_mount SetLocaleHook
```
hackartisan, line 7: old comment?

lib/dpul_collections_web/live/home_live.ex
```diff
@@ -21,15 +24,18 @@ defmodule DpulCollectionsWeb.HomeLive do
...
           </div>
         </form>
       </div>
       <div id="welcome" class="grid place-self-center gap-10 max-w-prose">
-        <h3 class="text-5xl text-center">Explore Our Digital Collections</h3>
+        <h3 class="text-5xl text-center"><%= gettext("Explore Our Digital Collections") %></h3>
         <p class="text-xl text-center">
-          We invite you to be inspired by our globally diverse collections of <%= @item_count %> Ephemera items. We can't wait to see how you use these materials to support your unique research.
+          <%= gettext("We invite you to be inspired by our globally diverse collections of") %> <%= @item_count %>
+          <%= gettext(
```
hackartisan, line 36: Does the break between tags render as a newline?

priv/gettext/default.pot
```diff
@@ -0,0 +1,130 @@
...
+msgstr ""
+
+#: lib/dpul_collections_web/components/core_components.ex:80
+#, elixir-autogen, elixir-format
+msgid "Success!"
+msgstr ""
+
+#: lib/dpul_collections_web/components/core_components.ex:85
+#, elixir-autogen, elixir-format
+msgid "We can't find the internet"
```
hackartisan, line 47: this one is my favorite

priv/gettext/es/LC_MESSAGES/default.po
```diff
@@ -0,0 +1,130 @@
...
+#: lib/dpul_collections_web/live/item_live.ex:61
+#: lib/dpul_collections_web/live/search_live.ex:151
+#, elixir-autogen, elixir-format
+msgid "Pages"
+msgstr "Páginas"
+
+#: lib/dpul_collections_web/live/search_live.ex:186
+#, elixir-autogen, elixir-format
+msgid "Previous"
+msgstr "anterior"
```
hackartisan, line 90: should this one be capitalized?

test/dpul_collections_web/live/home_live_test.exs
```diff
@@ -6,7 +6,7 @@ defmodule DpulCollectionsWeb.HomeLiveTest do
   test "GET /", %{conn: conn} do
     count = DpulCollections.Solr.document_count()
     conn = get(conn, ~p"/")
-    assert html_response(conn, 200) =~ "#{count} Ephemera items"
+    assert html_response(conn, 200) =~ "#{count}\n      Ephemera items"
```
hackartisan, line 9: I think it would be better if we could not have this newline. Relevant to my earlier comment

test/dpul_collections_web/set_locale_hook_test.exs
```diff
@@ -0,0 +1,58 @@
+defmodule DpulCollectionsWeb.SetLocaleHookTest do
+  use ExUnit.Case, async: true
+  # import Phoenix.LiveView
```
hackartisan, line 3: old?

hackartisan (changes requested):
This looks great!!! I left a few comments.

Also I'd love to look together, I don't understand why we have `set_locale_hook` as well as `locale_plug` and how they interact. I'm hoping you can show me.

lib/dpul_collections_web/set_locale_hook.ex
```diff
@@ -0,0 +1,13 @@
+defmodule DpulCollectionsWeb.SetLocaleHook do
```
hackartisan, line 1: Could we add a module doc that says something like, "set up locale for front-end use"

lib/plug/locale_plug.ex
```diff
@@ -0,0 +1,99 @@
+defmodule DpulCollectionsWeb.LocalePlug do
+  @moduledoc """
+  Fetch locale and set it for the specified Gettext backend.
+
```
hackartisan, line 4: maybe add something like, "sets it in the session so the front end can use it"

priv/gettext/en/LC_MESSAGES/default.po
```diff
@@ -0,0 +1,130 @@
...
+
+#: lib/dpul_collections_web/live/search_live.ex:186
+#, elixir-autogen, elixir-format
+msgid "Previous"
+msgstr ""
+
+#: lib/dpul_collections_web/live/search_live.ex:69
+#, elixir-autogen, elixir-format, fuzzy
+msgid "Search Results"
+msgstr "Search"
```
hackartisan, line 95: Remove this one

lib/dpul_collections_web/components/lux-components.ex
sdellis: I think it's fine to leave it. It would be fairly easy to add the Brasilian translations if we wanted to make sure that option works.

lib/dpul_collections_web/live/home_live.ex
sdellis: It does. The text is very long. I usually have soft wrapping on in my editor, but I try to avoid lines that are that long. I am not tied to having a line break here.

hackartisan: There's a max width style on it so I don't think it will ever get too long on the actual screen.

test/dpul_collections_web/set_locale_hook_test.exs
hackartisan: @sdellis one more

hackartisan (approved):
🎉

PR #256, opened 2024-11-20
Conditionally render link or span in pagenation

### LINKED ISSUE #168: Update paginator logic to not render "..." separator as a link.
Follow on to https://github.com/pulibrary/dpul-collections/pull/149.

## Acceptance Criteria

- [ ] Hovering over the `...` in the paginator doesn't look like a link, and continues not doing anything.
- [ ] There shouldn't be an anchor tag around the `...`

## Technical Info

This might be hard. It's also possible to do with CSS, but that would leave an anchor around. If we leave an anchor around, we'd have to add something so screen readers will ignore it.

The "..." paginator separators are currently rendered as links which do not trigger paginate events:

https://github.com/pulibrary/dpul-collections/blob/36c66f631f081e9ac192468b6d3be8387765bf1c/lib/dpul_collections_web/live/search_live.ex#L209-L210

https://github.com/pulibrary/dpul-collections/blob/36c66f631f081e9ac192468b6d3be8387765bf1c/lib/dpul_collections_web/live/search_live.ex#L172

It would be better if these were rendered as a span or div or some other element.

## First Step

Write a test.

hackartisan:
closes #168

### DIFF
--- a/lib/dpul_collections_web/live/search_live.ex
+++ b/lib/dpul_collections_web/live/search_live.ex
@@ -1,16 +1,15 @@
 defmodule DpulCollectionsWeb.SearchComponents do
   use Phoenix.Component
 
-  attr :rest, :global, doc: "the arbitrary HTML attributes to add to the flash container"
   attr :text, :string, doc: "the page number, or ellipsis"
-  attr :element, :string, doc: "'a' or 'span'"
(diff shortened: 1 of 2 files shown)


eliotjordan (approved):
Nice! 🚀

Issue #268, opened 2024-11-22
Track down missing EphemeraFolders items in dpul-c


eliotjordan:
The numer of EphemeraFolders in solr do not match the number of open and complete EphemeraFolders in Figgy.

Items in Figgy from UI: 56,164
Items in Figgy from query: 56,174
HydrationCache: 56,105
TransformationCache: 56,105
Items in Solr:    55,651

tpendragon:
@eliotjordan Do the number of hydration cache entries match?

eliotjordan:
@tpendragon I'm working on that number now. The number of transformation cache entries does not match, though. I'll post updated stats in a sec.

eliotjordan:
Ids in transformation cache that are missing in solr index:
[missing_solr_ids.csv](https://github.com/user-attachments/files/17875466/missing_solr_ids.csv)

Loggin of solr responses in dpul-c show that items with blank titles cause errors in the solr slug generator script. Theory is that the whole batch of solr records is not indexed if it contains a record without a title. 

From Graphana:
```
2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F     "msg":"Unable to invoke function processAdd in script: slug.js: TypeError: null has no such function \"replace\" in <eval> at line number 9", |  
-- | -- | --
  |   | 2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F       "root-error-class","org.openjdk.nashorn.internal.runtime.ECMAException"], |  
  |   | 2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F       "error-class","org.apache.solr.common.SolrException", |  
  |   | 2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F     "metadata":[ |  
  |   | 2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F   "error":{ |  
  |   | 2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F     "QTime":52}, |  
  |   | 2024-11-22 13:36:14.420 | 2024-11-22T13:36:14.137187548-06:00 stdout F     "status":500, |  
```

eliotjordan:
Grafana logs for Solr records that raised an error:

[...]

PR #269, opened 2024-11-22
Handle transformation and solr indexing errors

eliotjordan:
Work towards #268

- Index documents one at a time after batch Solr update returns an error
- Assign default value when title is missing
- Resolves a couple of test warnings.

Current counts:
Figgy: 56172
Hydration Cache: 56106
Transformation Cache: 56106
Solr Records: 56106

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
@@ -27,7 +27,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
 
     %{
       id: id,
-      title_txtm: get_in(metadata, ["title"]),
+      title_txtm: extract_title(metadata),
       description_txtm: get_in(metadata, ["description"]),
       years_is: extract_years(data),
       display_date_s: format_date(metadata),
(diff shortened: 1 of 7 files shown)


lib/dpul_collections/solr.ex
```diff
@@ -96,11 +98,35 @@ defmodule DpulCollections.Solr do
...
+
+  def add(docs, collection) when length(docs) > 1 do
+    response =
+      Req.post!(
+        update_url(collection),
+        json: docs
+      )
+
+    if response.status != 200 do
+      Enum.each(docs, fn doc -> add([doc]) end)
```
tpendragon, line 111: Suggestion: `docs |> Enum.each(&add/1)` or `Enum.each(docs, &add/1)`

lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
```diff
@@ -27,7 +27,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
 
     %{
       id: id,
-      title_txtm: get_in(metadata, ["title"]),
+      title_txtm: extract_title(metadata),
```
tpendragon, line 30: I like that we're starting to get to a pattern here.

tpendragon (approved):
Just a small style suggestion, but no blockers.

Looks great!

lib/dpul_collections/solr.ex
eliotjordan: Had to change the guard clauses a bit, but overall, I think it's better now.

tpendragon (approved):

PR #310, opened 2025-02-13
Get all members at once, rather than one at a time from the db

### LINKED ISSUE #227: Member hydration should use a members query instead of querying for every resource
## Reasoning

Right now if a folder has 600 pages, it hydrates by sending 600 postgres queries. We can make that 1.

## Acceptance Criteria

- [ ] Getting members for hydration takes only 1 query.

## First step

Look at https://github.com/pulibrary/figgy/blob/ex_figgy/ex_manifest/apps/ex_manifest_api/lib/ex_manifest_api/resource/query.ex#L6-L21 for inspiration, where we gathered up members for an old Elixir-based manifest generation service.

hackartisan:
closes #227

### DIFF
--- a/test/dpul_collections/indexing_pipeline_test.exs
+++ b/test/dpul_collections/indexing_pipeline_test.exs
@@ -99,6 +99,19 @@ defmodule DpulCollections.IndexingPipelineTest do
       assert IndexingPipeline.get_figgy_resource!(ephemera_folder_id).id == ephemera_folder_id
     end
 
+    test "get_figgy_resource!/1 errors when the resource isn't there" do
+      ephemera_folder_id = "00000000-0000-0000-0000-000000000000"
+
+      assert_raise Ecto.NoResultsError, fn ->
+        IndexingPipeline.get_figgy_resource!(ephemera_folder_id)
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


test/dpul_collections/indexing_pipeline_test.exs
```diff
@@ -99,6 +99,19 @@ defmodule DpulCollections.IndexingPipelineTest do
       assert IndexingPipeline.get_figgy_resource!(ephemera_folder_id).id == ephemera_folder_id
     end
 
+    test "get_figgy_resource!/1 errors when the resource isn't there" do
```
tpendragon, line 102: Where'd this test come from?

tpendragon (approved):

tpendragon:
Just one question, but approved - feel free to merge.

test/dpul_collections/indexing_pipeline_test.exs
hackartisan: I added this, it seemed like the error case wasn't tested and then I used it as a model to check the behavior of the new function. Do you think it's redundant?

tpendragon: Not at all, it just seemed unrelated so I was curious if there was something in developing this that brought a bug to light.

PR #317, opened 2025-02-14
Delete records linked to deletion markers

eliotjordan:
- **Delete solr documents linked to deletion markers**
- **Remove hydration and transformation cache entries for deleted items**

### DIFF
--- a/architecture-decisions/0004-deleting-records.md
+++ b/architecture-decisions/0004-deleting-records.md
@@ -1,0 +1,60 @@
+# 4. Deleting Records
+
+Date: 2025-02-18
+
+## Status
+
+Accepted
+
+## Context
+
+When resources are deleted in Figgy, a DeletionMarker resource is created at the
+same time. The DeletionMarker stores the deleted resource's identifier,
+resource type, and a serialized copy of the  metadata (in the `deleted_object`
+field). We need a method for processing DeletionMarkers in DPUL-C to remove the
+corresponding record from the Solr index.
+
+## Decision
+
+#### Hydration Consumer
+
+1. We will process DeletionMarkers that reference a deleted resource with a
+resource type that we currently index into DPUL-C. In addition, we will check
+if a hydration cache entry exists for the deleted resource and discard the 
+DeletionMarker if not.
+1. A special CacheMarker is created from the DeletionMarker that uses the
+   deleted resource's id as the id and the updated_at value from the
+   DeletionMarker as the timestamp.
+1. Special hydration cache entry attributes are generated. The hydration cache
+   entry created from these attributes will replace the hydration cache entry of
+   the deleted resource.
+    -  Existing metadata is replaced with a simple deleted => true kv pair
+    -  The entry id is set to the deleted resource's id
+    -  The entry internal_resource type is set to that of the deleted resource
+
+#### Transformation Consumer
+
+1. Messages with the deleted => kv pair are handled separately.
+1.A special solr document is generated from the deleted object hydration cache
+entry with the following structure.
+  ```
+    %{ id: "id", deleted: true }
+  ```
+
+####  Indexing Consumer
+
+1. Messages with the `deleted: true` field are handled sperately and assigned to
+   the `delete` batcher.
+1. The delete batcher sends the deleted record ids to a the Solr.delete_batch
+   function which iterates over them, deletes each record, and then commits the
+   batch of deletes.
+
+## Consequences
+
+- DeletionMarkers will stay in the Figgy database unless the resource is
+restored. This means that DPUL-C will have to triage an ever increasing number
+over time.
+
+- Deleted resource hydration and transformation cache entries will stay in the 
+cache after the resource is remove from Solr until the next full reindex, or in
+the case of the transformation cache, partial reindex.
--- a/architecture-decisions/0004-deleting-records.md
+++ b/architecture-decisions/0004-deleting-records.md
@@ -34,8 +34,7 @@ DeletionMarker if not.
 
 #### Transformation Consumer
 
-1. Messages with the deleted => kv pair are handled separately.
-1.A special solr document is generated from the deleted object hydration cache
+1. A special solr document is generated from the deleted object hydration cache
 entry with the following structure.
   ```
     %{ id: "id", deleted: true }
@@ -47,7 +46,9 @@ entry with the following structure.
    the `delete` batcher.
 1. The delete batcher sends the deleted record ids to a the Solr.delete_batch
    function which iterates over them, deletes each record, and then commits the
-   batch of deletes.
+   batch of deletes. The additional batcher doesn't create a potential race
+   condition because there is only one entry for the resource earlier in the
+   pipeline.
 
 ## Consequences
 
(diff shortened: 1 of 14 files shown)


eliotjordan:
Tested in staging. Deleting a test EphemeraFolder in Figgy caused the corresponding record to be removed from DPUL-C.

lib/dpul_collections/indexing_pipeline/figgy/hydration_consumer.ex
```diff
@@ -88,6 +88,43 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationConsumer do
...
+          }
+        },
+        %{cache_version: cache_version}
+      )
+      when internal_resource in ["DeletionMarker"] and
+             resource_type in ["EphemeraFolder", "EphemeraTerm"] do
+    # Only process messages where the deleted resource has an exisiting
+    # hydration cache entry. If one does not exist, it means that the resource
+    # has not been indexed into DPUL-C.
+    resource = IndexingPipeline.get_hydration_cache_entry!(resource_id, cache_version)
```
tpendragon, line 109: Should we be passing this cache_version? I think this is the `cache_version` for the DeletionMarker, no?

lib/dpul_collections/indexing_pipeline/figgy/resource.ex
```diff
@@ -65,6 +72,23 @@ defmodule DpulCollections.IndexingPipeline.Figgy.Resource do
...
   @spec to_map(resource :: %__MODULE__{}) :: map()
+  defp to_map(resource = %__MODULE__{internal_resource: "DeletionMarker"}) do
+    %{"resource_id" => [%{"id" => deleted_resource_id}], "resource_type" => [resource_type]} =
+      resource.metadata
+
+    # Create attributes specifically for deletion markers.
+    # 1. Replace existing metadata with a simple deleted => true kv pair
+    # 2. Set the entry id to the deleted resource's id
+    # 3. Set the entry internal_resource type to that of the deleted resource
+    resource
```
tpendragon, line 83: Suggestion: maybe we can just explicitly create the map here? Unless there's still some properties we're saving.

lib/dpul_collections/solr.ex
```diff
@@ -149,6 +149,20 @@ defmodule DpulCollections.Solr do
...
+  def delete_batch(ids, collection \\ read_collection()) do
+    ids
+    |> Enum.each(fn id ->
+      Req.post!(
+        update_url(collection),
+        json: %{delete: %{query: "id:#{id}"}}
+      )
+    end)
+
+    commit(collection)
```
tpendragon, line 163: I like committing, I wonder if we need a soft commit? Otherwise we're looking at a potentially long write lock.

tpendragon (changes requested):
This is really great, just a couple questions.

lib/dpul_collections/indexing_pipeline/figgy/hydration_consumer.ex
```diff
@@ -88,6 +88,43 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationConsumer do
...
+
+    cond do
+      resource ->
```
hackartisan, line 112: call this `entry`?

tpendragon, line 112: Suggestion: Change variable name to `hydration_cache_marker`

tpendragon (changes requested):

lib/dpul_collections/indexing_pipeline/figgy/transformation_consumer.ex
```diff
@@ -49,6 +49,24 @@ defmodule DpulCollections.IndexingPipeline.Figgy.TransformationConsumer do
   @impl Broadway
   @spec handle_message(any(), any(), %{required(:cache_version) => integer()}) ::
           Broadway.Message.t()
+  def handle_message(
```
hackartisan, line 52: We think we can remove this, the function is the same it's just the document gets created differently

architecture-decisions/0004-deleting-records.md
```diff
@@ -0,0 +1,60 @@
...
+1. Special hydration cache entry attributes are generated. The hydration cache
+   entry created from these attributes will replace the hydration cache entry of
+   the deleted resource.
+    -  Existing metadata is replaced with a simple deleted => true kv pair
+    -  The entry id is set to the deleted resource's id
+    -  The entry internal_resource type is set to that of the deleted resource
+
+#### Transformation Consumer
+
+1. Messages with the deleted => kv pair are handled separately.
```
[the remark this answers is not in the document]
hackartisan: Or are they?

architecture-decisions/0004-deleting-records.md
```diff
@@ -0,0 +1,60 @@
...
+entry with the following structure.
+  ```
+    %{ id: "id", deleted: true }
+  ```
+
+####  Indexing Consumer
+
+1. Messages with the `deleted: true` field are handled sperately and assigned to
+   the `delete` batcher.
+1. The delete batcher sends the deleted record ids to a the Solr.delete_batch
```
hackartisan, line 48: can we add something about how the additional batcher doesn't create a race condition because there's only one entry for the resource earlier in the pipeline?

lib/dpul_collections/solr.ex
hackartisan: Write a follow-up ticket to optimize this deletion, we think maybe there's a delete handler we could use to do them all at once

hackartisan (changes requested):

lib/dpul_collections/indexing_pipeline/figgy/hydration_consumer.ex
tpendragon: I got cache_version confused with cache_order.

tpendragon (approved):

lib/dpul_collections/indexing_pipeline/figgy/transformation_consumer.ex
```diff
@@ -49,6 +49,24 @@ defmodule DpulCollections.IndexingPipeline.Figgy.TransformationConsumer do
...
+
+    message
+    |> Message.put_data(%{
+      marker: marker,
+      incoming_message_data: hydration_cache_entry,
+      handled_data: solr_doc
+    })
+  end
+
   def handle_message(
```
hackartisan: Since they're so similar, it would be nice to have a comment saying why we are keeping both. I'm thinking it's so we can filter out things that aren't ephemera folders?

hackartisan (changes requested):
One last thing!

lib/dpul_collections/indexing_pipeline/figgy/transformation_consumer.ex
eliotjordan: @hackartisan I previously removed that transformation consumer function. No changes made to this file in the PR currently.

hackartisan: weird, browser stuff I guess. Okay!

hackartisan (approved):

PR #318, opened 2025-02-17
Can we have different executor for the format task

hackartisan:
that doesn't run all the containers?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,10 @@ executors:
       - image: pulibrary/ci-solr:8.4-v1.0.0
         command: server/scripts/ci-start.sh
     working_directory: ~/project
+  formatter:
+    docker:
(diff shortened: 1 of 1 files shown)


tpendragon:
Was the format task running slow or something?

hackartisan:
> Was the format task running slow or something?

I was waiting for it to finish when the test had already passed at one point last week, and that annoyed me so I started messing with this. I don't really know if the stats in general support it.

hackartisan:
This run seems within the bounds of the recent history of formatting check runs, so maybe it's not any faster. but it's at the low end of that bound. I can try rerunning it, too, to get another sample.

hackartisan:
Ran this a second time and it was a few seconds faster than the previous fastest run.

tpendragon:
We might be able to get faster in the future by caching the compile or something, but for now this is great.

PR #322, opened 2025-02-21
Set up wallaby and add a feature test

### LINKED ISSUE #311: Set Up Wallaby for Client-Side Feature Tests
### Summary or User Story
`Phoenix.LiveView.JS` functions are used when we want client-side interactions without making a trip to the server. However, there is no way to test these with LiveViewTest. Setting up Wallaby will allow us to write feature tests to ensure client-side code is working as expected.

### Acceptance Criteria

This section can be completed by the DLS team. Some requirements to consider
including:

- [ ] Using a `JS` function such as `JS.toggle` is testable
- [ ] New functionality is documented

### First step

Follow the README at https://github.com/elixir-wallaby/wallaby

hackartisan:
Followed the instructions in the wallaby readme, using chromdriver
because that seems to be the most common setup that works well for
people, based on looking around the elixir forum

I added a new mix alias for running tests in ci because you have to
ensure that js is built otherwise liveview stuff doesn't work. there may
be a day when we need to also add the css assets?

Finally, I skipped everything in the setup about running tests async and
ensuring database connections don't mess with one another. We may need
to do that stuff at some point.

Note that failing feature tests will add a screenshot artifact

closes https://github.com/pulibrary/dpul-collections/issues/311


tpendragon (approved):
Was able to get this to run locally, looks great!

PR #328, opened 2025-02-27
Browse Design Pass #1

sdellis:
Here's a screenshot, but there is some interaction transitions with the browse button. 

Screenshot 2025-02-27 at 12 14 01 PM


sdellis:
Interact here: https://dpul-collections-staging.lib.princeton.edu/browse

PR #333, opened 2025-03-06
Improve language selection by setting cookie with current locale

### LINKED ISSUE #265: Selecting a language should preserve other querystring parameters
### Summary or User Story

If I submit a search and look at browse results, then select "spanish" from the language dropdown, I stay on the search page but I lose my search results and the page number I was on

### Acceptance Criteria

- [ ] Selecting a language from the dropdown preserves other querystring parameters

### First step

write a test

### implementation tips

this may be helpful https://elixirforum.com/t/how-do-you-build-urls-in-your-views-while-maintaining-query-params/57114/2

eliotjordan:
Instead of relying on a locale parameter to set language, set enable the language dropdown to set the locale cookie and reload. Avoids removing parameters from url when changing language.

Closes #265

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -33,6 +33,15 @@ topbar.config({barColors: {0: "#29d"}, shadowColor: "rgba(0, 0, 0, .3)"})
 window.addEventListener("phx:page-loading-start", _info => topbar.show(300))
 window.addEventListener("phx:page-loading-stop", _info => topbar.hide())
 
+// Event to change locale cookie from Language drop down and reload page so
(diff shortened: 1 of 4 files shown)


tpendragon:
I think the only downside here is that this is the only part of the site that would fall apart without JS on, but maybe that's just reality.

tpendragon:
> I think the only downside here is that this is the only part of the site that would fall apart without JS on, but maybe that's just reality.

lol no this is not true, I just disabled JS, nothing works, it just kinda looks like it might.

tpendragon (approved):

PR #337, opened 2025-03-07
Allow browse items to be pinned to the top.

### LINKED ISSUE #315: Add a way to "float up" or mark certain favorite items in the browse display
### Summary or User Story

There should be a button on each item which will re-order that item to the front of the search results and give it some kind of visual indicator that it's been stuck there.

### Acceptance Criteria

- [ ] There's a button on each item, when pushed it adds it to the pinned area
- [ ] Items that have been pinned stay in the "random" area but have some kind of visual indicator that it's been pinned
- [ ] Items in the "pinned" section displays more thumbnails like our search results
- [ ] clicking the "random" button retains the pinned items
- [ ] There's a sticky floating button that takes me to my pins and has a count of how many pinned items there are that I can see increment when I click the pin icon. (this could be done as a follow-up PR)

### First step

Write a test.

tpendragon:
There's a bunch of design work to do here, but this gets the basic "popping to the top" working.

Closes #315 

Screen Shot 2025-03-25 at 12 23 58 PM

Screen Shot 2025-03-25 at 12 24 05 PM

### DIFF
--- a/lib/dpul_collections_web/live/browse_live.ex
+++ b/lib/dpul_collections_web/live/browse_live.ex
@@ -38,7 +38,7 @@ defmodule DpulCollectionsWeb.BrowseLive do
       ) do
         doc = items |> Enum.find(fn item -> item.id == id end)
     if idx = Enum.find_index(pinned_items, fn pinned_item -> doc.id == pinned_item.id end) do
-      socket = socket |> assign(pinned_items: List.delete_at(pinned_items, idx)) |> dbg
+      socket = socket |> assign(pinned_items: List.delete_at(pinned_items, idx))
       {:noreply, socket}
     else
       {:noreply, socket |> assign(pinned_items: Enum.concat(pinned_items, [doc]))}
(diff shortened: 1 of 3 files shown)


lib/dpul_collections_web/live/browse_live.ex
```diff
@@ -29,8 +33,31 @@ defmodule DpulCollectionsWeb.BrowseLive do
...
   end
 
+  def handle_event(
+        "pin",
+        %{"item_id" => id},
+        socket = %{assigns: %{items: items, pinned_items: pinned_items}}
+      ) do
+    doc = items |> Enum.find(fn item -> item.id == id end)
+
+    if idx = Enum.find_index(pinned_items, fn pinned_item -> doc.id == pinned_item.id end) do
```
eliotjordan, line 43: discussion: Not blocking, but I am wondering if `case` would be more idiomatic here since we're assigning the variable idx. Something like:
```
case Enum.find_index(pinned_items, fn pinned_item -> doc.id == pinned_item.id end) do
  nil ->
      {:noreply, socket |> assign(pinned_items: Enum.concat(pinned_items, [doc]))}
  idx ->
      socket = socket |> assign(pinned_items: List.delete_at(pinned_items, idx))
      {:noreply, socket}
end

eliotjordan (commented):
Looks good! Feel like we can iterate more on the design as we add functionality (e.g. "you might also like")

lib/dpul_collections_web/live/browse_live.ex
tpendragon: Thank you for the suggestion! I was very unsure.

eliotjordan (approved):

PR #350, opened 2025-03-17
Request smaller images for non-primary thumbnails

### LINKED ISSUE #347: Request smaller images for non-primary thumbnails in browse
Requesting smaller images that are closer to the rendered size for non-primary thumbnail images in the browse results will reduce overall traffic and image load times.

eliotjordan:
Closes #347


sdellis (approved):
LGTM 👍

PR #367, opened 2025-03-19
Add more metadata

### LINKED ISSUE #341: Index more metadata into solr document
### Summary or User Story

We have a lot of fields in the database that should be easy to index, let's just put them in `_txtm` fields so they're searchable.

### Acceptance Criteria

The following fields are indexed (do as many as are relatively easy):

- [x] alternative_title
- [x] sort_title
- [x] transliterated_title
- [x] contributor
- [x] creator
- [x] publisher
- [x] date_created
- [x] date_range
- [x] height
- [x] width
- [x] barcode
- [x] folder_number
- [x] keywords
- [x] content_warning
- [x] description
- [x] holding_location
- [x] provenance
- [x] series

For each field: 

- [x] Each search field can be searched
- [x] They're on the item record page. Don't worry about the design of it, they're just there.
- [x] There's a ticket created for fields that aren't straight forward.

### First step

Add some fields to https://github.com/pulibrary/dpul-collections/blob/main/test/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry_test.exs#L10-L28.

## Technical Implementation

Instead of writing a test for every field, we should write a test that checks that the output of solrizing the fixtures is equivalent to a given hard-coded map.

eliotjordan:
Closes #341 

Screenshot 2025-03-19 at 4 15 02 PM

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
@@ -45,7 +45,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
       title_txtm: extract_title(metadata),
       alternative_title_txtm: get_in(metadata, ["alternative_title"]),
       barcode_txtm: get_in(metadata, ["barcode"]),
-      content_warning_txtm: get_in(metadata, ["content_warning"]),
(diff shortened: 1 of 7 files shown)


tpendragon:
Screen Shot 2025-03-19 at 3 22 47 PM
Apparently I'm in dark mode :joy:

Issue #368, opened 2025-03-19
Index nested or related fields into Solr


eliotjordan:
### Summary or User Story

We know that nested fields or values from related parent resources will require special hydration and indexing to handle. At the end of this, those fields should all be indexed in `_txtm` fields and searchable.

### Acceptance Criteria

Fields:

- [x] box_number
- [x] rights statement

### First step

Write a test.

PR #377, opened 2025-03-20
Index rights statement

eliotjordan:
Work towards #368


tpendragon (approved):
We might need those URIs someday, but I think this is great for now.

PR #389, opened 2025-03-25
Index box_number from parent EphemeraBox

### LINKED ISSUE #368: Index nested or related fields into Solr

eliotjordan:
Closes #368 

Screenshot 2025-03-25 at 1 04 11 PM


lib/dpul_collections/indexing_pipeline/figgy/resource.ex
```diff
@@ -93,7 +107,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.Resource do
   defp to_map(
          resource = %__MODULE__{metadata: %{"state" => [state], "visibility" => [visibility]}}
        )
-       when state != "complete" or visibility != "open" do
+       when state not in ["complete", "all_in_production"] or visibility != "open" do
```
tpendragon, line 110: OH, interesting. Presumably converting the box hit this because the box has state? I think we still want the box number even if it's not all in production or complete. I wonder if this guard should be more strict to just EphemeraFolders?

lib/dpul_collections/indexing_pipeline.ex
```diff
@@ -238,6 +238,25 @@ defmodule DpulCollections.IndexingPipeline do
     |> FiggyRepo.all()
   end
 
+  def get_figgy_parents(id) do
```
tpendragon, line 241: I do wonder how hard it would be to do a query for parents like we do in Figgy. Would you mind if I gave it a shot?

tpendragon (changes requested):
Just a couple small thoughts. Exciting to have box numbers in!

lib/dpul_collections/indexing_pipeline.ex
tpendragon: https://github.com/pulibrary/dpul-collections/pull/391 PR of what that looks like.

eliotjordan: Merged.

lib/dpul_collections/indexing_pipeline/figgy/resource.ex
eliotjordan: Done

lib/dpul_collections/indexing_pipeline.ex
tpendragon, line 241: Looks like https://github.com/pulibrary/dpul-collections/pull/391 got lost in a rebase?

tpendragon (changes requested):

tpendragon (approved):

PR #390, opened 2025-03-25
updates random values from 3000 to 1000000 for the item browse

AminEdZare:
Question before merging:

I noticed that I could modify the URL in the following way: "/browse?r=1000abc" and results would still be generated (no button press involved).

image

Depending on how result sets are currently stored on the server-side, I was wondering if allowing for this functionality enables a malicious user to consume lots of memory in the event they repeatedly automatically generate their own queries. (e.g.) "/browse?r=1000abcd", "/browse?r=1000abcde", "/browse?r=1000abcdee" ...

If this is the case, should a ticket for cleaning the input and ensuring only numbers between the specified range can actually be queried be made?

### DIFF
--- a/lib/dpul_collections_web/live/browse_live.ex
+++ b/lib/dpul_collections_web/live/browse_live.ex
@@ -21,12 +21,12 @@ defmodule DpulCollectionsWeb.BrowseLive do
 
       {:noreply, socket}
     else
-      {:noreply, push_patch(socket, to: "/browse?r=#{Enum.random(1..1000000)}", replace: true)}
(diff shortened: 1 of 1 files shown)


tpendragon:
> Depending on how result sets are currently stored on the server-side, I was wondering if allowing for this functionality enables a malicious user to consume lots of memory in the event they repeatedly automatically generate their own queries.

Good question. It's stored for the current session until they disconnect, but should clean itself up when they navigate to a new page, or if they never connect to the websocket. If we want to clean that up in the future we'll probably migrate to Streams.

tpendragon (changes requested):
Looks good, just needs a `mix format`

tpendragon (approved):

PR #396, opened 2025-03-28
Fetch and display related controlled vocabulary terms

eliotjordan:
- **Add multi query**
- **Fetch and display related controlled vocabulary terms**
- **Extract FileSet members from related_data resources instead of separate list**


lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
```diff
@@ -88,14 +93,14 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
 
   defp primary_thumbnail_service_url(
          %{"thumbnail_id" => thumbnail_id} = metadata,
-         %{"member_ids" => member_data} = related_data
```
tpendragon, line 91: I don't think this is a problem right now, but one potential thing with the members change is that IN() in postgres has a max of ~32,000 entries. I think in our cases that's probably fine (I can't imagine a 32k page book, and I doubt a project is going to have 32k boxless folders), but good to be aware that now there's a technical limit.

eliotjordan: Added note.

tpendragon (changes requested):
Just a small suggestion. I really love how compact and generic this solution is!

lib/dpul_collections/indexing_pipeline/figgy/resource.ex
```diff
@@ -53,25 +53,29 @@ defmodule DpulCollections.IndexingPipeline.Figgy.Resource do
...
     }
   end
 
-  @spec extract_members(resource :: %__MODULE__{}) :: related_resource_map()
-  defp extract_members(resource = %{:metadata => %{"member_ids" => _member_ids}}) do
-    # Enum.reduce(member_ids, %{}, &append_related_resource/2)
-    # get just the list of members
-    # turn it into a map of id => FiggyResource
-    IndexingPipeline.get_figgy_members(resource.id)
+  defp fetch_related(%__MODULE__{metadata: metadata}) do
```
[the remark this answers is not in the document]
eliotjordan: Comments added.

lib/dpul_collections/indexing_pipeline/figgy/resource.ex
```diff
@@ -53,25 +53,83 @@ defmodule DpulCollections.IndexingPipeline.Figgy.Resource do
...
     }
   end
 
-  @spec extract_members(resource :: %__MODULE__{}) :: related_resource_map()
-  defp extract_members(resource = %{:metadata => %{"member_ids" => _member_ids}}) do
-    # Enum.reduce(member_ids, %{}, &append_related_resource/2)
-    # get just the list of members
-    # turn it into a map of id => FiggyResource
-    IndexingPipeline.get_figgy_members(resource.id)
+  @doc """
```
tpendragon, line 61: LOVE it.

lib/dpul_collections/indexing_pipeline/figgy/resource.ex
```diff
@@ -53,25 +53,83 @@ defmodule DpulCollections.IndexingPipeline.Figgy.Resource do
...
+            "id" => %{"id" => "0cff895a-01ea-4895-9c3d-a8c6eaab4017"},
+            "internal_resource" => "FileMetadata",
+            "mime_type" => ["image/tiff"],
+            "use" => [%{"@id" => "http://pcdm.org/use#ServiceFile"}]
+          }
+        ]
+      }
+    }
+  """
+  @spec fetch_related(%__MODULE__{}) :: map()
```
tpendragon, line 104: Suggestion: Non-blocking, but there's a `related_data()` type in this file, maybe that can be used as the return value in this typespec.

tpendragon (approved):

lib/dpul_collections/indexing_pipeline/figgy/resource.ex
eliotjordan: Good idea. Fixed.

tpendragon (approved):

PR #397, opened 2025-03-28
Extract parent from related_data resources instead of separate list

eliotjordan:
Proposal to not use separate query just for parent. How sure are we that `cached_parent_id` has issues? This is a simpler option.

### DIFF
--- a/lib/dpul_collections/indexing_pipeline.ex
+++ b/lib/dpul_collections/indexing_pipeline.ex
@@ -207,14 +207,6 @@ defmodule DpulCollections.IndexingPipeline do
   """
   def get_figgy_resource!(id), do: FiggyRepo.get!(Figgy.Resource, id)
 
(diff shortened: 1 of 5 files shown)


tpendragon:
> How sure are we that `cached_parent_id` has issues?

That's a good question. On save in Figgy the parent gets cached here: https://github.com/pulibrary/figgy/blob/main/app/change_set_persisters/change_set_persister/cache_parent_id.rb.

To find out how sure I am I'd have to run a report of every resource, get its parent, and compare it to the cached_parent_id and tell me if there wasn't a match. I could probably do that, if you think the simplicity here is worth it?

Another thing to consider is `cached_parent_id` is definitely not authoritative - in Figgy we only use it for an OAI endpoint, as described here: https://github.com/pulibrary/figgy/issues/3917

PR #421, opened 2025-04-04
Toggles sticky toolbar visibility with intersection observer

AminEdZare:


sdellis:
Beyond formatting issues and failing test, there is a bug in this code that @AminEdZare and I noticed: 

When the sticky tools are visible and the user pins an item, it pushes the `pin` event to the server. The sticky_tools component gets re-rendered, setting the opacity back to zero. We should preserve the opacity on re-render, perhaps with a life-cycle hook?

AminEdZare:
We would need a ticket to handle the accessibility of tabbing through the sticky menu, which is currently not working as tab just navigates down the resource list.

assets/js/hooks.js
```diff
@@ -0,0 +1,25 @@
+let Hooks = {};
+
+Hooks.ToolbarHook = {
+  mounted() {
+
+    const sideMenu = document.getElementById("sticky-tools");
```
tpendragon, line 6: What's this sideMenu used for?

assets/js/hooks.js
```diff
@@ -0,0 +1,25 @@
...
+
+    this.observer = new IntersectionObserver(entries => {
```
tpendragon, line 8: So my understanding is that `this` here is the browse header, since that's what has the phx-hook attached, yeah? Awesome!

lib/dpul_collections_web/live/browse_live.ex
```diff
@@ -50,17 +51,29 @@ defmodule DpulCollectionsWeb.BrowseLive do
     end
   end
 
+  def handle_event("show_stickytools", _params, socket) do
+    {:noreply, update(socket, :show_stickytools?, fn _val -> true end)}
```
tpendragon, line 55:
```suggestion
    {:noreply, assign(socket, :show_stickytools?, true)}
```

lib/dpul_collections_web/live/browse_live.ex
```diff
@@ -50,17 +51,29 @@ defmodule DpulCollectionsWeb.BrowseLive do
...
+  end
+
+  def handle_event("hide_stickytools", _params, socket) do
+    {:noreply, update(socket, :show_stickytools?, fn _val -> false end)}
```
tpendragon, line 59:
```suggestion
    {:noreply, assign(socket, :show_stickytools?, false)}
```

tpendragon (changes requested):
Just a couple small changes. This looks and works really great, good work @AminEdZare !

assets/js/hooks.js
AminEdZare: Not used for anything, left from a prior implementation approach, cleaned up!

assets/js/hooks.js
AminEdZare: Yes!

lib/dpul_collections_web/live/browse_live.ex
AminEdZare: changed!

lib/dpul_collections_web/live/browse_live.ex
AminEdZare: done

tpendragon (approved):

tpendragon (approved):

PR #422, opened 2025-04-04
Hydration Consumer should only use small messages while possible.

### LINKED ISSUE #420: HydrationConsumer handle_messages should be responsible for querying the full resource, not the producer
### Summary (Product Owner)

The HydrationProducerSource should only produce enough information to filter in the Consumer, but not the entire record - because generating the entire record for Consumer*Demand is very expensive, since it has to parse all the metadata. Another benefit is we won't use as much memory copying these resources from the producer process to the consumer processes.

It should get the ID, timestamp, state (metadata->'state' AS state), internal_resource, and visibility

### Implementation Notes

You might want to name the thing that gets returned something.

Don't SELECT all of metadata, you want whatever the Ecto version of "SELECT metadata->'state' AS state" is. Maybe ecto will handle this somehow if you only select metadata->'state'? But if not, we might not be able to use Figgy.Resource.

### Acceptance Criteria

- HydrationProducerSource produces these smaller records, but Consumer queries for the full record if it isn't filtered out.

### First Step

Write a test

tpendragon:
Closes #420

I'm not sure if this is my favorite implementation - sometimes there's
metadata, and sometimes there's not, and it keys off of that knowledge
to know if it should query for more, and I don't totally love that. It does work though!

### DIFF
--- a/test/dpul_collections/indexing_pipeline/figgy/hydration_consumer_test.exs
+++ b/test/dpul_collections/indexing_pipeline/figgy/hydration_consumer_test.exs
@@ -12,10 +12,8 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationConsumerTest do
           id: "47276197-e223-471c-99d7-405c5f6c5285",
           updated_at: ~U[2018-03-09 20:19:34.486004Z],
           internal_resource: "EphemeraFolder",
-          metadata: %{
-            "state" => ["complete"],
-            "visibility" => ["open"]
-          }
+          state: ["complete"],
+          visibility: ["open"]
         }
       }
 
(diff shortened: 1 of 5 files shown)


test/dpul_collections/indexing_pipeline/figgy/hydration_consumer_test.exs
```diff
@@ -86,9 +78,12 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationConsumerTest do
           file_set_deletion_marker_message
         ]
         |> Enum.map(&Figgy.HydrationConsumer.handle_message(nil, &1, %{cache_version: 1}))
+
+      message_batchers =
+        transformed_messages
         |> Enum.map(&Map.get(&1, :batcher))
 
-      assert transformed_messages == [:default, :noop, :noop, :default, :noop, :noop]
+      assert message_batchers == [:default, :noop, :noop, :default, :noop, :noop]
```
hackartisan, line 86: 👍🏻

hackartisan (approved):
I like it.

PR #424, opened 2025-04-04
Track cache versions for indexing metrics.

### LINKED ISSUE #419: Track cache version as part of index metric
### Summary or User Story

When more than one index is running we can't rely on the index tracker because it doesn't distinguish between events coming from different cache version processes.

### Acceptance Criteria

- [ ] the index metrics tracker genserver keeps separate state for a processor running more than once b/c there are multiple cache versions
- [ ] the dashboard shows which cache version a report row is for 

### First step

This section can be completed by the DLS team

tpendragon:
Closes #419

### DIFF
--- a/lib/dpul_collections/index_metrics_tracker.ex
+++ b/lib/dpul_collections/index_metrics_tracker.ex
@@ -97,7 +97,7 @@ defmodule DpulCollections.IndexMetricsTracker do
     state =
       state
       |> put_in(
-        [Access.key(source.processor_marker_key(), %{}), Access.key(cache_version, %{})],
+        [Access.key(processor_marker_key, %{}), Access.key(cache_version, %{})],
         handle_ack_received(metadata, get_in(state, [processor_marker_key, cache_version]))
       )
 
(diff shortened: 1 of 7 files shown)


priv/repo/migrations/20250404213147_add_cache_version_to_index_metric.exs
```diff
@@ -0,0 +1,11 @@
+defmodule DpulCollections.Repo.Migrations.AddCacheVersionToIndexMetric do
+  use Ecto.Migration
+
+  def change do
+    alter table("index_metrics") do
+      add :cache_version, :integer, default: 1
```
hackartisan, line 6: should we default nil? I think '1' would potentially be a lie?

hackartisan (changes requested):
Yay! Comments as per pair review

lib/dpul_collections/index_metrics_tracker.ex
```diff
@@ -97,7 +97,7 @@ defmodule DpulCollections.IndexMetricsTracker do
     state =
       state
       |> put_in(
-        Enum.map([processor_marker_key, cache_version], &Access.key(&1, %{})),
+        [Access.key(source.processor_marker_key(), %{}), Access.key(cache_version, %{})],
```
hackartisan, line 100: ah, this one has extra `source`, making the test fail

hackartisan (changes requested):

hackartisan (approved):

Issue #409, opened 2025-04-01
Implement home page mockup


tpendragon:
### Summary or User Story

We've decided that our home page mockup is far enough along that future iteration can happen with CSS rather than in Illustrator, so we're going to implement the design!

Follow on to #344 

### Mock-up

Insert mock up here after the final touch ups.

### Acceptance Criteria

It's often difficult to create a design straight from a mockup image, and it's expected that not everything will be 1:1. If you find there are some elements that are particularly challenging, you're encouraged to create follow up tickets.

- [ ] The mock up above is implemented for the home page.

### Technical Tips

You can find the images used in the mock-up (and other options) at https://docs.google.com/document/d/184ldOFhXhJYNv_Z04M2qaS2dJX6XpFV71MLl-TNk75w/edit?tab=t.0

Tips for typography and other things might go here, insert that info if you know it.

### First step

Pick an element of the mock up and rev up the CSS!

sdellis:
## Typography

_Note: Do not use Google Fonts for loading. Download and use in the app._

### Headline / Display Font: Poppins (ExtraBold and Bold)
[Download Here](https://fonts.google.com/specimen/Poppins)
Used for headings and large type.

### San-serif Prose Font: Poppins (Regular)
[Download Here](https://fonts.google.com/specimen/Poppins)
Used for smaller text blocks and descriptions.

### Serif Prose Font: Aleo 
[Download Here](https://fonts.google.com/specimen/Aleo)
Used (italicized) for callouts, asides, and quotes. Also used (regular) for long-form text (essays and longer articles) for  improved readability.  
(Aleo is not present in this homepage mockup, but it was in the triangle one. I wanted to keep the typography stuff together so I included it here.)

## Color Palette 

- Taupe: #ede7e4
- Pink: #e2d4c9
- Sage Green: #d0d3ac
- Cloud Blue: #c1ccc3
- Rust / Orange: #ab6b47
- Dark Blue: #3a4151

![Image](https://github.com/user-attachments/assets/577b39e2-2fd4-4b38-a3f7-3ba21c56f790)

tpendragon:
[...]

PR #431, opened 2025-04-09
Adds fixtures to test font support for arabic and cyrillic characters

sdellis:
refs #409


sdellis:
FYI - This container has already been built and pushed.

tpendragon (approved):

PR #442, opened 2025-04-16
Moving version of Mosaic

tpendragon:
Option for #441 

https://github.com/user-attachments/assets/b27b6329-2dd3-453a-9726-dd88437eb6fa


tpendragon:
We felt like the movement was a lot, we're gonna close this and try some other things.

Issue #437, opened 2025-04-15
Implement home page content area designs


hackartisan:
### Summary or User Story

Reference #409 to implement the "explore ephemera" and "recently added items" areas of the home page design

spun off from #409 in attempt to keep the #430 branch from running even longer

### Acceptance Criteria

This section can be completed by the DLS team. Some requirements to consider
including:

- [x] Explore ephemera area is implemented
- [ ] Recently added items area has been updated to match design

### First step

write some HTML and CSS

PR #444, opened 2025-04-16
Add recently added to sort_by

tpendragon:
This also does some refactoring of sort_by, so to add new things we just update `Solr.Constants`. `https://hexdocs.pm/elixir/module-attributes.html#as-compile-time-constants` said that's pretty common. Maybe it should be elsewhere though? I dunno.

The macro business is a little complicated - basically we have to have module attributes to use them in function guards, but you can't call attributes from another module, so we have to compile-time define them in places we want to use those constants. To do that we can do `use` and `using`, effectively working as a module include from Ruby (hmmm, maybe that's bad..), to make it available to the compiler.

We also have to call Gettext's macro directly, because if I `use Gettext` in that `__using__` then it complains for modules that already `use Gettext`. If we instead call `gettext` with dynamic values, then GetText is unable to recognize those strings for the mix tasks, and I like that workflow.

Work supporting #437

### DIFF
--- a/lib/dpul_collections/solr.ex
+++ b/lib/dpul_collections/solr.ex
@@ -1,5 +1,6 @@
 defmodule DpulCollections.Solr do
   require Logger
+  use DpulCollections.Solr.Constants
 
   @spec document_count(String.t()) :: integer()
   def document_count(collection \\ read_collection()) do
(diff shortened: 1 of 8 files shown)


hackartisan (approved):
I like it

PR #451, opened 2025-04-18
Improve Figgy fixture behavior

### LINKED ISSUE #449: Need a method for adding new fixtures to local container without pushing to registry
Adding new fixtures tends to break the full integration test because of resource counts used to signal when indexing is complete. We need a script that will update our local container without pushing to the GitHub registry and breaking tests for everyone.

eliotjordan:
- Enable loading "synthetic" fixtures from CSV files.
- Stops building custom fixture image and use stock Postgres container instead
- Load fixture records from command line / mix task.

Closes #449


tpendragon:
I get the idea, seems good. Do we want to keep directions on how to generate those binary files at all?

eliotjordan:
@tpendragon I think we want to keep the directions for now. We can revisit in the future. I updated the fixture instructions just a bit.

README.md
```diff
@@ -34,14 +34,49 @@ Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
...
+- `cd figgy-fixture-container && ./create-fixture-exports.sh`
 
-To add a new fixture, edit `create-fixture-exports.sh` and add an id to the
-EXTRA_RESOURCE_IDS var. Note that adding a new fixture will update all the
-existing and you may have to make adjusts to test expectations.
+ To add extra fixtures:
+- Edit `figgy-fixture-container/create-fixture-exports.sh` and add an id to the EXTRA_RESOURCE_IDS var
+- `cd figgy-fixture-container && ./create-fixture-exports.sh`
+
+Note that refreshing or adding a new fixture might require yout to make adjusts to test expectations.
```
hackartisan, line 47:
```suggestion
Note that refreshing or adding a new fixture might require you to make adjustments to test expectations.
```

hackartisan (changes requested):
I like this. Just found one line with typos. I will also try it out.

README.md
eliotjordan: Good catch! Fixed.

tpendragon (approved):

hackartisan (approved):

PR #454, opened 2025-04-21
Index non-complete EphemeraFolders if parent is complete

### LINKED ISSUE #382: Ephemera Folders should index if their box is "all_in_production"
### Summary or User Story

Right now we only accept "complete" Ephemera Folders, but ephemera folders also go live if their boxes are "All In Production." This is likely why we don't see a number of items that are live in DPUL.

### Acceptance Criteria

- [ ] A folder that is "needs_qa", but in a box that's "all_in_production" should index.

### First step

Write a test.

eliotjordan:
Closes #382 

- **Send all EphemeraFolders to the default batcher**
- **For non-complete EphemeraFolders check index if parent EphemeraBox is complete**


test/dpul_collections/indexing_pipeline/integration/full_integration_test.exs
```diff
@@ -21,18 +21,29 @@ defmodule DpulCollections.IndexingPipeline.FiggyFullIntegrationTest do
...
     # this into account.
     deleted_resource_count = FiggyTestSupport.deletion_marker_count()
-    # Resources that are not public or not complete
-    non_indexable_resource_count = 4
-    total_records = ephemera_folder_count + deleted_resource_count - non_indexable_resource_count
+    total_records = ephemera_folder_count + deleted_resource_count
     # Empty resources are resources with no image file sets
     empty_resource_count = 1
 
+    if System.get_env("DEBUG_INTEGRATION") do
```
eliotjordan, line 28: I'm not sure if we should keep this, but this is an area where I'm constantly inserting the same IO.puts statements. In the long term, I think we need a better (or more centralized) way to calculate what the expected counts are.

tpendragon: If it's workin' for ya, happy to keep it around. Maybe this is a place we should add telemetry?

tpendragon (approved):
This looks great to me!

As an aside, I'd really like to sit down with your experience some time and see if we can get a pattern in place for this kind of thing - I'm finding myself falling out of context already in these reviews, and I wonder if there's some sort of pipeline design we can do to separate some of these concepts of filtering, enriching, deleting, etc now that we've got most of the logic for folders in place.

PR #464, opened 2025-04-23
Search design pass

sdellis:
WIP... 
Currently fulfills acceptance criteria but needs tests. Blocked on making genre search actionable, but that can be a separate ticket. 

Notes: 
1. Pills are not removable via the button/icon. 
Q: How should this behave? Similar to Blacklight pills where only half the element (with the close icon) is clickable? Should the whole thing be clickable to reset values?

2. Blank search messaging... is the wording on this pill okay? (Blank search terms returning * seems to only be a thing in libraries.)

### Screenshot
Screenshot 2025-04-23 at 4 16 28 PM

### Mobile View
Screenshot 2025-04-23 at 4 16 07 PM

### DIFF
--- a/test/support/test_utils.ex
+++ b/test/support/test_utils.ex
@@ -1,11 +1,11 @@
 defmodule TestUtils do
-    def clean_string(string) do
+  def clean_string(string) do
     # Replaces non-breaking space with regular space
     # Collapses all whitespace into a single space
     # Trims leading/trailing spaces
     string
-        |> String.replace("\u00A0", " ")
-        |> String.replace(~r/\s+/, " ")
-        |> String.trim()
-    end
+    |> String.replace("\u00A0", " ")
+    |> String.replace(~r/\s+/, " ")
+    |> String.trim()
+  end
 end


tpendragon:
> 1. Q: How should this behave? Similar to Blacklight pills where only half the element (with the close icon) is clickable? Should the whole thing be clickable to reset values?

Let's try making the whole thing a button I think?

> Blank search messaging... is the wording on this pill okay? (Blank search terms returning * seems to only be a thing in libraries.)

It's fine. I'd also be okay with not having a pill for the search term, and just having that in the top like it was, but either way.

sdellis:
@tpendragon thank you for the comments! 

I actually think that it's pretty important to show the "state" of a blank search because it is not always intuitive. At the very least, it should show in the `<H1>` at the top ("Search Results for: [all possible items]"). 

However, what happens when you remove that pill? If it resets `q` to blank, the pill becomes permanent and you can't remove it. Yeah, I think I will remove that one!

test/dpul_collections_web/features/locale_test.exs
```diff
@@ -30,4 +31,22 @@ defmodule DpulCollectionsWeb.Features.LocaleTest do
 
     Solr.delete_all(active_collection())
   end
+
+  test "renders facet with states", %{conn: conn} do
```
hackartisan, line 35: Can we move this test to the search_live_test file?

hackartisan (changes requested):
This looks really good, I have two questions

- can we make the "added time ago" text only display if it's sorting by recently added? I don't think it's super relevant info otherwise?

- On the browse cards I ended up removing the number of pages if there were less than the maximum the card could show. What do you think of doing that here? So it would only show how many pages if there were 5 or more pages.

test/dpul_collections_web/features/locale_test.exs
sdellis: Hmm... @hackartisan that was the only way I could test the conditionals for whether to show or hide the pill element. Is there a better way to test this?

sdellis: Oh, I see what you mean... move it out of the locale test. Yes! It should go in the `features/search_bar_test.exs` file, right?

sdellis:
> This looks really good, I have two questions
> 
>     * can we make the "added time ago" text only display if it's sorting by recently added? I don't think it's super relevant info otherwise?
> 
>     * On the browse cards I ended up removing the number of pages if there were less than the maximum the card could show. What do you think of doing that here? So it would only show how many pages if there were 5 or more pages.

All of this makes sense. Will do!

tpendragon (changes requested):
Just a couple small suggestions for comment. This looks great!

lib/dpul_collections_web/components/browse_item.ex
```diff
@@ -70,7 +70,7 @@ defmodule DpulCollectionsWeb.BrowseItem do
       </div>
       
     <!-- "added on" note -->
-      <div :if={@added?} class="self-end w-full bg-taupe h-10 p-2 text-right">
+      <div :if={@added?} class="digitized_at self-end w-full bg-taupe h-10 p-2 text-right">
```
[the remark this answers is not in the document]
sdellis: That is so I can query it and test the feature here: 

https://github.com/pulibrary/dpul-collections/pull/464/files#diff-237552747f7b64b2ecac71e4166be58b8a57845a6f9df0110557281a9db188d4R53-R55

lib/dpul_collections_web/components/layouts/app.html.heex
```diff
@@ -1,5 +1,5 @@
 <.live_component module={DpulCollectionsWeb.SearchBarComponent} id="search-bar" />
-<main id="main-content" class="bg-taupe page-y-padding">
+<main id="main-content" style="min-height: calc(100vh - 18rem);" class="bg-taupe page-y-padding">
```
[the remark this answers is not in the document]
sdellis: Hmm... maybe! I will look into it.

tpendragon:
@sdellis I think I accidentally blew up your test I'm sorry :(

hackartisan:
> With regard to the browse card pages, I don't think that approach will work correctly on mobile due to images hidden by css overflow. Perhaps we should make this into its own ticket?

Yes, that sounds good to me

lib/dpul_collections_web/components/layouts/app.html.heex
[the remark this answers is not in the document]
sdellis: @tpendragon I have been working on this for a while and can't seem to get it to work with flex. 

On top of that, I cannot get Tailwind to work for this CSS rule: `min-height: calc(100vh - 18rem)`. It should be doable with `min-h-[calc(100vh - 18rem)]` but it doesn't work. That's why I added it as an inline style. 

I was able to add the `taupe` background color to the wrapper in `root.html.heex` and it seems to do the trick. LMK if this works for you.

```
<div class="flex-1 bg-taupe">
        {@inner_content}
</div>
```

test/dpul_collections_web/features/locale_test.exs
sdellis: @hackartisan yes, I was able to move it into search_live_test.

lib/dpul_collections_web/live/search_live.ex
```diff
@@ -98,7 +99,14 @@ defmodule DpulCollectionsWeb.SearchLive do
     ~H"""
     <div class="content-area">
       <h1>
-        {gettext("Search Results for")}: <span class="normal-case">{@search_state.q}</span>
+        {gettext("Search Results for")}:
+        <span class="normal-case">
+          <%= if @search_state.q do %>
+            {@search_state.q}
+          <% else %>
+            [ {gettext("All Possible Items")} ]
```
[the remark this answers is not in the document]
sdellis: This is now extracted to its own function.

priv/gettext/default.pot
```diff
@@ -53,38 +59,38 @@ msgstr ""
 msgid "Language"
 msgstr ""
 
-#: lib/dpul_collections_web/live/search_live.ex:255
+#: lib/dpul_collections_web/live/search_live.ex:311
 #, elixir-autogen, elixir-format
 msgid "Next"
 msgstr ""
 
+<<<<<<< HEAD
```
tpendragon, line 67: Looks like just a rebase miss. Maybe just re-running the translation tasks would help?

tpendragon (changes requested):
Just a small rebase thing. It works really well!

test/dpul_collections_web/features/search_bar_test.exs
```diff
@@ -28,4 +28,17 @@ defmodule DpulCollectionsWeb.Features.SearchBarTest do
     |> click_button("Search")
     |> assert_has("#item-counter", text: "1 - 1 of 1")
   end
+
+  test "renders facet with states", %{conn: conn} do
```
hackartisan, line 32: Could this test go in `search_live_test`? It's really testing the search results page, as opposed to the search bar component.

hackartisan (changes requested):

hackartisan (approved):

PR #473, opened 2025-04-25
Action Bar Content

### LINKED ISSUE #471: Record page: lay out and style action bar
spun off from https://github.com/pulibrary/dpul-collections/issues/433; see that ticket for mockup

tpendragon:
Closes #471 

Screen Shot 2025-04-25 at 2 58 47 PM

Screen Shot 2025-04-25 at 3 48 16 PM

I've spun out the share and size button functionality into their own tickets #476 and #475 

## Missing Content

This won't display rights images for Cicognara Microfiche or Museum of FIne Arts, Boston rights, because they don't have images. I don't have a good fall back for when an image isn't found. Fortunately neither of those are selected for Ephemera.


hackartisan (approved):
This looks great! I wonder if we'll want these to be links to CC or rightsstatements.org? If so, future ticket.

Issue #75, opened 2024-08-22
Incremental updates of data stored by reference


tpendragon:
Hydrator should support incremental updates of dependent rows for Ephemera Folders

## Acceptance Criteria

- [ ] If a Genre is updated, every folder that uses that genre should rehydrate to include that genre's new information.

## Technical Tips

When a related item appears in the hydration consumer, query for all records in the cache that store that item and have a cache_order less than the related resource's cache_order. Rehydrate those.

Add a source_cache_order_record_id to keep track of which record the cache_order came from.

Update the cache_order when you hydrate a record to be the latest timestamp between the record and all related resources.

## First Step

Look at the documentation below.

tpendragon:
Conversation w/ @eliotjordan 

![Image](https://github.com/user-attachments/assets/352b0b1b-4105-4e75-8380-f87bddef844c)

The key takeaway is that the `source_cache_order` should, at hydration time, be the timestamp of the latest timestamp between the record and all related data.

PR #480, opened 2025-04-29
Index EpheremaProject title

### LINKED ISSUE #461: Index Ephemera Project Title into Solr
### Summary or User Story

Folders should have their project title indexed.

Folders can either be in a box which is in a project, or in a project directly.

There's only ever one Project, so it can be indexed in a single valued field.

### Acceptance Criteria

- [x] Ephemera projects are hydrated.
- [ ] Ephemera project titles are indexed in the Ephemera Folder document.

### First step

Write a test.

eliotjordan:
Closes #461 

- **Hydrate EphemeraProjects**
- **Hydrate EphemeraBoxes**
- **Recursively fetch ancestors**
- **Get parent EphemeraBox from ancestors map**
- **Ensure parent resource isn't fetched twice**

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
@@ -40,21 +40,22 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
         data: data = %{"metadata" => metadata},
         related_data: related_data
       }) do
-    parent_metadata = extract_parent_metadata(metadata, related_data)
+    box_metadata = extract_box_metadata(related_data)
+    project_metadata = extract_project_metadata(related_data)
 
     %{
       id: id,
       title_txtm: extract_title(metadata),
       alternative_title_txtm: get_in(metadata, ["alternative_title"]),
       barcode_txtm: get_in(metadata, ["barcode"]),
-      box_number_txtm: get_in(parent_metadata, ["box_number"]),
+      box_number_txtm: get_in(box_metadata, ["box_number"]),
       content_warning_txtm: get_in(metadata, ["content_warning"]) |> remove_empty_strings,
       contributor_txtm: get_in(metadata, ["contributor"]),
       creator_txtm: get_in(metadata, ["creator"]),
       description_txtm: get_in(metadata, ["description"]),
       digitized_at_dt: digitized_date(data),
       display_date_s: format_date(metadata),
-      ephemera_project_title_s: extract_project_title(related_data),
+      ephemera_project_title_s: Map.get(project_metadata, "title", []) |> Enum.at(0),
       file_count_i: file_count(metadata),
       folder_number_txtm: get_in(metadata, ["folder_number"]),
       genre_txtm: extract_term("genre", metadata, related_data),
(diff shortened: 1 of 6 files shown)


lib/dpul_collections/indexing_pipeline/figgy/hydration_cache_entry.ex
```diff
@@ -171,16 +173,46 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationCacheEntry do
...
   defp extract_service_url(nil), do: nil
 
-  defp extract_parent_metadata(%{"cached_parent_id" => [%{"id" => cached_parent_id}]}, %{
-         "parent_ids" => parent_data
-       }) do
-    parent_data
-    |> get_in([cached_parent_id])
-    |> get_in(["metadata"])
+  defp extract_box_metadata(%{"ancestors" => ancestors}) when map_size(ancestors) > 0 do
+    box = find_ancestor_type(ancestors, "EphemeraBox")
```
tpendragon, line 177: Love this!

lib/dpul_collections/indexing_pipeline/figgy/hydration_consumer.ex
```diff
@@ -135,7 +135,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.HydrationConsumer do
         %{cache_version: cache_version}
       )
       when internal_resource in ["DeletionMarker"] and
-             resource_type in ["EphemeraFolder", "EphemeraTerm"] do
+             resource_type in ["EphemeraFolder", "EphemeraTerm", "EphemeraProject", "EphemeraBox"] do
```
tpendragon, line 138: Why are we hydrating boxes and folders?

tpendragon (changes requested):
My understanding is:

1. At hydration of an EphemeraFolder, recursively query for parents and add each parent as a flat list of ancestors.
2. At transformation time, find the ancestor box and folder and use those for indexing.

I love that, I think it's pretty simple and will work well.

I also noticed that you're now hydrating boxes and projects (I think?) What are those hydrated boxes/projects used for?

lib/dpul_collections/indexing_pipeline/figgy/hydration_consumer.ex
eliotjordan: It was an early iteration of the PR that I decided to leave in place. I'm thinking that it might be necessary for https://github.com/pulibrary/dpul-collections/issues/75. EphemeraTerms are also hydrated. We're leaving them in for the same reason, I suppose.

tpendragon: @eliotjordan I don't think we're going to need it for #75. I don't think we need Terms either. Should we talk about it synchronously so I can understand better where you're comin' from?

eliotjordan: It's OK. I'll remove it and make a ticket for removing hydration of Terms.

eliotjordan: https://github.com/pulibrary/dpul-collections/issues/481

tpendragon (approved):

PR #498, opened 2025-05-23
Genre Facet

### LINKED ISSUE #485: Allow faceting by genre (no UI)
### Summary or User Story

follow-up to #464 and #433 

The record page "genre" value should be a clickable link that takes you to a search result filtered to that genre.

### Acceptance Criteria

- [ ] genre values can be applied and removed as filters
- [ ] clicking the genre in a record page takes you to the results page with that filter applied
- [ ] clicking the (x) pill for that genre successfully removes the genre filter from the results list

out of scope: a UI in the search results page to filter to any genre

### First step

Check the genre field in the index to ensure it's a facetable field, if it's not probably create a copy field for the facet. write a test.

tpendragon:
Closes #485 

This refactors existing facets to be much more configurable (in `Constants.ex`) and makes faceting work in a much more general way, to make it easier to add simple facets later.

It also converts faceting to use the `fq` parameter (which Solr can cache, and which is a little easier to combine separate from the query param)

Screen Shot 2025-06-02 at 4 05 12 PM

### DIFF
--- a/lib/dpul_collections/solr.ex
+++ b/lib/dpul_collections/solr.ex
@@ -32,7 +32,8 @@ defmodule DpulCollections.Solr do
 
     solr_params = [
       q: query_param(search_state),
-      "q.op": "AND",
+      mm: "6<90%",
+      fq: facet_param(search_state),
       fl: fl,
       sort: sort_param(search_state),
       rows: search_state[:per_page],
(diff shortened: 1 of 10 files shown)


sdellis (approved):
The code is very readable and more flexible for future facets. I tried it out in the browser and it works great. Nice work! 👍

PR #502, opened 2025-05-28
Even closer on the browse callout images

hackartisan:
This iteration on the browse callout zooms even closer on each image to evoke playful curiosity and emphasize the collections over the individual items

Screenshot 2025-05-28 at 13-00-30 Digital Collections · Phoenix Framework


tpendragon:
@hackartisan I think this implementation will mean the cache will always miss. I wonder if we can do this with CSS?

Issue #488, opened 2025-05-02
Design pass: colors


hackartisan:
### Summary or User Story

We may like to try some other colors.

### Acceptance Criteria

- [ ] Any member of the team who would like to propose alternate color schemes has done so to their satisfaction, and the team has discussed those proposals

One or more PRs may address this issue. Work may or may not be merged as a result of this ticket.

PR #503, opened 2025-05-28
Color Iteration: Monochrome Search Bar

tpendragon:
This proposal tries to reduce the number of different colors in the palette by making the secondary section a shade of the search bar. The goal is to maintain serenity, but have the color profile feel more consistent.

**Before**
Screen Shot 2025-05-28 at 10 40 45 AM

**After**
Screen Shot 2025-05-28 at 10 40 48 AM

Work towards #488

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -24,6 +24,21 @@
   /* Button Text on Dark Blue */
   --color-light-green: #ECF2CB;
 
+  /* Some semantic colors */
+  --color-background: var(--color-taupe);
+  --color-light-text: var(--color-taupe);
+  --color-dark-text: var(--color-dark-blue);
+  /* brand is the color that's the site's brand - header/footer */
+  --color-brand: var(--color-dark-blue);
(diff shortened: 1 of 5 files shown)


tpendragon:
> 1. I like the color scheme of version 1 better, but im still hung up on that lighter spring green to the left.

I'm reading this as "I like the blue colors more than the sage colors in general", is that right?

kelea99:
I guess so? I also wonder what colors work best in a search bar in contrast with the text you are typing in to search, or the text you have there in slightly darker/similar colors. Is there some color/colors that one should not use due to things like color blindness, is there? I maybe overthinking now.

sdellis:
I agree that this is a step in the right direction, although I don't think it's quite there yet. I would be fine with merging this, but to continue to refine it. Agree with Kim that the Search Bar color is not quite right, maybe too much hue? I don't have a lot of confidence in providing alternative color/tint/hue suggestions without trying them out myself first.

PR #507, opened 2025-05-29
Lighten the colors on the home page in particular, use sage as primary

hackartisan:
Oh, this also uses the princeton black as the brand color, and integrates that color into the site via the text color.

refs #488

After
Screenshot 2025-05-29 at 10-56-49 Digital Collections · Phoenix Framework

Screenshot 2025-05-29 at 10-47-09 Digital Collections · Phoenix Framework

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -24,17 +24,20 @@
   /* Button Text on Dark Blue */
   --color-light-green: #ECF2CB;
 
+  --color-princeton-black: #121212;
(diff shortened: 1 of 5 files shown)


tpendragon (approved):
I think this is a step in the right direction, the black header/footer is a little intense but in the context of the wider brand will be an easier fit, and the changes to the structure make it a lot easier to adjust.

Good work.

PR #508, opened 2025-05-29
Colors: Try out Canva Sage.

tpendragon:
This maintains the sage theme, but uses a slightly hue shifted set of colors that come from Canva. It also standardizes some text and fixes a couple of strange button behaviors to use accent instead of the background color.

Work towards #488

Screen Shot 2025-05-29 at 9 08 28 AM
Screen Shot 2025-05-29 at 9 08 34 AM
Screen Shot 2025-05-29 at 9 08 40 AM


assets/css/app.css
```diff
@@ -42,14 +52,19 @@
    * Images/highlights/hyperlinks/boxes/cards/etc.
    */
   --color-accent: var(--color-rust);
-  --color-search: var(--color-sage);
+  --color-search: var(--color-sage-200);
+  --color-light-green: var(--color-dark-text);
```
hackartisan, line 56: this is confusing why is light-green being set to the value of dark-text?

hackartisan (changes requested):

PR #509, opened 2025-05-29
Display PDF Button

### LINKED ISSUE #287: Only show pdf download link if resource allows / provides pdf
### Summary or User Story

Some resources don't have a pdf download, so they should not present that as an option. This means we'll need to do some logic when we have the whole figgy resource, i.e. at index time

example record: 1dffc2e2-84ce-4775-93a4-d282d6bf2fc5

follow-up to #257 / #286

### Acceptance Criteria

This section can be completed by the DLS team. Some requirements to consider
including:

- [ ] When a record can't generate a PDF because it's disabled, the button should be disabled and say "No PDF Available"

### Technical Tips

We might have to replicate some logic in Figgy to make this call.

Might be able to find it by searching Figgy for "No PDF available for this item"

### First step

Write a test. We'll have to figure out the logic and then replicate a fixture.

tpendragon:
Closes #287

This also fixes the semantics for links/buttons in the Item page by
creating links when there's an `href`, and buttons otherwise.

Screen Shot 2025-05-29 at 10 51 18 AM

### DIFF
--- a/lib/dpul_collections_web/live/item_live.ex
+++ b/lib/dpul_collections_web/live/item_live.ex
@@ -198,7 +198,7 @@ defmodule DpulCollectionsWeb.ItemLive do
   def download_button(assigns) do
     ~H"""
     <.primary_button disabled>
-      No PDF Available
+          {gettext("No PDF Available")}
     </.primary_button>
     """
   end
--- a/lib/dpul_collections_web/live/item_live.ex
+++ b/lib/dpul_collections_web/live/item_live.ex
@@ -178,17 +178,28 @@ defmodule DpulCollectionsWeb.ItemLive do
         height="800"
       />
 
-      <.primary_button class="left-arrow-box" href="#" target="_blank">
+      <.primary_button class="left-arrow-box">
         <.icon name="hero-eye" /> {gettext("View")}
       </.primary_button>
 
-      <.primary_button
-        href={"#{Application.fetch_env!(:dpul_collections, :web_connections)[:figgy_url]}/catalog/#{@item.id}/pdf"}
-        target="_blank"
-      >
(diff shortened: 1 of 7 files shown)


lib/dpul_collections_web/live/item_live.ex
```diff
@@ -235,15 +246,24 @@ defmodule DpulCollectionsWeb.ItemLive do
...
         {render_slot(@inner_block)}
-      </a>
+      </div>
+    </a>
+    """
+  end
+
+  def primary_button(assigns) do
+    ~H"""
+    <button class={["btn-primary flex gap-2", @class]} disabled={@disabled}>
```
sdellis, line 265: Should a `disabled` button look different from `btn-primary`? Maybe a grey with a `cursor: not-allowed` style? 

Also, based on the test, it seems that one (the only?) scenario for unavailable PDFs is that the user does not have permissions. It might be helpful, maybe in a future PR to indicate to the user that they may have to log in to download the PDF.

sdellis (commented):
This looks good. Just a suggestion about better distinguishing the disabled button from the primary-button.

lib/dpul_collections_web/live/item_live.ex
tpendragon: > Also, based on the test, it seems that one (the only?) scenario for unavailable PDFs is that the user does not have permissions. It might be helpful, maybe in a future PR to indicate to the user that they may have to log in to download the PDF.

I don't think we have a scenario where someone can't access the item, but can download the PDF if they login. Which is to say, if the PDF is set to "none" then even authed users can't get it, unless they're like Figgy staff.

tpendragon: Anna made this follow up ticket: https://github.com/pulibrary/dpul-collections/issues/510

sdellis (approved):

Issue #433, opened 2025-04-09
Implement Record Page layout


tpendragon:
### Summary or User Story

The mockup developed in #370 is implemented here.

Blocked by #370, and blocked by #409

### Acceptance Criteria

- [ ] The UI designed and agreed on in #370 is implemented in real life.

If there is any particular part of the design implementation that's particularly challenging, folks should feel comfortable creating another ticket for implementing that part.

### First step

Start CSS'ing.

sdellis:
These are ROUGH numbers for the sizing of page elements and columns. We will need to make adjustments. 

### Column Size 
(For 'xl' screens / 1280px / 80 rem)
Left Column: 580px / 36 rem (fixed)
Right Column: 700px / 44 rem (fluid, not to exceed 80 chars)

### Fonts
Poppins - Titles, Headlines, Metadata 
Aleo - Descriptions and Long-form text 
Noto Nastaliq Urdu - Arabic/Urdu Scripts

H1 Font Size: 2rem
Text Font Size: 1rem

### Colors
Color Palette

    Taupe: #ede7e4
    Wafer Pink: #e2d4c9
    Sage Green: #d0d3ac
    Cloud Blue: #c1ccc3
    Rust / Orange: #ab6b47
    Dark Blue: #3a4151

![Image](https://github.com/user-attachments/assets/befb712b-34d8-4184-a44b-d3c0c63d41c8)


### Final Mockups

![Image](https://github.com/user-attachments/assets/54d1e9bc-6ccd-4404-9595-a99d16b1a547)

![Image](https://github.com/user-attachments/assets/64aaeb96-1df8-4da3-aac6-8ca6a1ce708d)

hackartisan:
Closing this since there's only 1 remaining sub-ticket

PR #517, opened 2025-06-03
Add more filters to item page

tpendragon:
Works as a way to prove the generic facet code works.

Screen Shot 2025-06-02 at 8 25 13 PM

Work towards #433

### DIFF
--- a/lib/dpul_collections/solr/constants.ex
+++ b/lib/dpul_collections/solr/constants.ex
@@ -42,7 +42,8 @@ defmodule DpulCollections.Solr.Constants do
         },
         "project" => %{
           solr_field: "ephemera_project_title_s",
-          label: Gettext.Macros.gettext_with_backend(DpulCollectionsWeb.Gettext, "Project"),
+          label:
+            Gettext.Macros.gettext_with_backend(DpulCollectionsWeb.Gettext, "Ephemera Project"),
           # Identity just returns whatever you gave it.
           value_function: &Function.identity/1
         },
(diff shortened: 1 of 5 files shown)


lib/dpul_collections/solr/constants.ex
```diff
@@ -39,6 +39,34 @@ defmodule DpulCollections.Solr.Constants do
           label: Gettext.Macros.gettext_with_backend(DpulCollectionsWeb.Gettext, "Genre"),
           # Identity just returns whatever you gave it.
           value_function: &Function.identity/1
+        },
+        "project" => %{
+          solr_field: "ephemera_project_title_s",
+          label: Gettext.Macros.gettext_with_backend(DpulCollectionsWeb.Gettext, "Project"),
```
hackartisan, line 45: I think I called it "Ephemera project" in the metadata pane. But maybe you're trying to be more terse here?

lib/dpul_collections_web/live/search_live.ex
```diff
@@ -171,7 +171,7 @@ defmodule DpulCollectionsWeb.SearchLive do
       navigate={self_route(@search_state, %{facet: %{@field => nil}})}
       class="filter mb-2 focus:border-3 focus:visible:border-rust focus:border-rust py-2 px-4 shadow-md no-underline rounded-lg bg-primary border-dark-blue text-white font-sans font-semibold text-sm btn-primary hover:text-white hover:bg-accent focus:outline-none active:shadow-none"
     >
-      {@label}
+      {Gettext.gettext(DpulCollectionsWeb.Gettext, @label)}
```
hackartisan, line 174: This is okay as long as they're spelled out somewhere else; maybe you could add a comment that says where in the code they are enumerated?

hackartisan (changes requested):
This is cool. I left a couple of comments please.

lib/dpul_collections/solr/constants.ex
tpendragon: Yeah it just seemed like a long facet label, but happy to keep it consistent for now.

PR #540, opened 2025-06-15
Add a "similarity" facet.

tpendragon:
I wondered if it'd be fun to browse the full "similarity" results.

### DIFF
--- a/lib/dpul_collections/solr.ex
+++ b/lib/dpul_collections/solr.ex
@@ -39,7 +39,8 @@ defmodule DpulCollections.Solr do
       fl: fl,
       sort: sort_param(search_state),
       rows: search_state[:per_page],
-      start: pagination_offset(search_state)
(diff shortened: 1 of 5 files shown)


lib/dpul_collections/solr.ex
```diff
@@ -40,7 +40,8 @@ defmodule DpulCollections.Solr do
       fl: fl,
       sort: sort_param(search_state),
       rows: search_state[:per_page],
-      start: pagination_offset(search_state)
+      start: pagination_offset(search_state),
+      uf: "* _query_"
```
hackartisan, line 44: Is this required to allow the embedded mlt query? Maybe we could add a comment here about why we have this value.

hackartisan (changes requested):
I rebased this and added one comment; I'd be interested in merging it.

PR #549, opened 2025-06-17
Image Counts are hidden in Search Result Row when all the resource's images are visible

sdellis:


lib/dpul_collections_web/live/search_live.ex
```diff
@@ -187,15 +187,25 @@ defmodule DpulCollectionsWeb.SearchLive do
...
       <div class="flex flex-wrap gap-5 md:max-h-60 max-h-[22rem] overflow-hidden justify-center md:justify-start relative">
         <.thumbs
           :for={{thumb, thumb_num} <- thumbnail_service_urls(5, @item)}
           :if={@item.file_count}
           thumb={thumb}
           thumb_num={thumb_num}
         />
-        <div :if={@item.file_count > 1} class="absolute right-0 top-0 bg-white px-4 py-2">
+        <div
+          :if={@item.file_count > 1}
```
tpendragon, line 205: I think if you delete this it'll fix the bug.

PR #554, opened 2025-06-24
Clicking thumbnail opens Clover to the corresponding Canvas

sdellis:
Adds href with viewer_url to the thumbnails so they open on the right canvas in Clover.


eliotjordan:
Closed in favor of https://github.com/pulibrary/dpul-collections/pull/560

PR #574, opened 2025-07-02
Cache thumbnails

### LINKED ISSUE #339: Cache thumbnails for faster rendering
### Summary or User Story

We've noticed in search results and the browse display that it takes a long time for images to generate from the IIIF server. We should try some techniques to improve that.

**Update**: Cloudfront is evicting our things every two days. We need a new strategy - this is what we're going to try:

In serverless-iiif we're going to accept a special header and when we get that header, we'll cache the tile to S3, much as it currently does for files that are too big to stream out of the lambda.

Also adjust the s3 bucket we're caching to so that it uses a lifecycle filter to never expire anything smaller than 1mb.

In the future we'll have to use a request function or something similar to do the same thing, because serverless-iiif no longer does that kind of failover.

### Acceptance Criteria

- [ ] The thumbnails load fast enough to feel "snappy" and joyful.

### First step

Choose an idea to try out.

## Technical Tips

We might be able to fetch every thumbnail in the background as part of our indexing process, or some other pipeline, so that Cloudflare caches those thumbnails for us.

eliotjordan:
Closes #339 

- **Install oban**
- **Add oban web dashboard**
- **Add CacheThumbnails worker**
- **Add worker to cache hero images**
- **Use Req to get images**
- **Add tests**
- **Use async tasks to run requests**
- **Stream request body**
- **Test request paths**
- **Schedule weekly caching of hero images**


eliotjordan:
TODO:

[ ] - Documentation
[ ] - Can we test the request parameters?
[ ] - Schedule hero image caching?
[ ] - Different ways to mock Req?

test/dpul_collections/workers/cache_thumbnails_test.exs
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    test "item thumbnails are cached when item does have primary thumbnail url" do
+      Oban.Testing.with_testing_mode(:inline, fn ->
+        doc = SolrTestSupport.mock_solr_documents(2) |> Enum.at(1)
+
+        # Setup ETS table to store image paths.
+        # Acts as storage for paths that are requested in different
+        # processes so they can be tested after the job has run.
+        :ets.new(:cached_paths, [:named_table, :public])
```
hackartisan, line 29: cool idea

config/prod.exs
```diff
@@ -22,5 +22,18 @@ config :logger, level: :info
...
+# Configures Oban job processing
+config :dpul_collections, Oban,
+  engine: Oban.Engines.Basic,
+  queues: [cache: 20],
+  repo: DpulCollections.Repo,
+  plugins: [
+    {Oban.Plugins.Cron,
+     crontab: [
+       # Cache Hero Images once a week. 1 AM every Monday.
+       {"0 1 * * MON", DpulCollections.Workers.CacheHeroImages, queue: :cache}
```
hackartisan, line 34: What is this use case for this schedule? If we were to add or change images I'd think we'd want to cache them at deploy.

architecture-decisions/0005-oban.md
```diff
@@ -0,0 +1,29 @@
...
+and we don't want to further slow our main indexing processes. In addition, we would
+like these image requests to retry if there are network or other issues.
+
+## Decision
+
+We will use [Oban](https://github.com/oban-bg/oban) to process asynchronous image caching jobs.
+
+## Consequences
+
+Adding Oban means we will have to periodically monitor it's queues like we do
```
hackartisan, line 25: typo: "its"

hackartisan (changes requested):
This looks good and it's so exciting how fast the images are. A couple of small comments.

architecture-decisions/0005-oban.md
eliotjordan: Fixed.

config/prod.exs
eliotjordan: I'm happy to remove this from a schedule, but I don't think caching every time we deploy is a great idea. That's a lot of unnecessary jobs/calls to the server. Think about how many times we might deploy in a week vs. how often these change. To me, this seemed like a good compromise. Maybe when we run start-over on the indexing consumer?

hackartisan: Oh, yeah, I also don't think we want to run this every time we deploy. Only when we deploy a change to the images! I guess if we kept them in a db table we could cache on each save or something? But yes a note in the code makes sense to me!!

eliotjordan: Done.

PR #592, opened 2025-07-14
Add accessibility testing

### LINKED ISSUE #173: Add automated accessibility testing
### Summary or User Story
Let's target WCAG AA. Automated testing can catch some problems, and we can periodically augment with manual testing.

### Acceptance Criteria

- [ ] Some automated testing tool runs as part of CI

### First step

search for tools

eliotjordan:
Closes #173

### DIFF
--- a/test/support/test_utils.ex
+++ b/test/support/test_utils.ex
@@ -11,13 +11,13 @@ defmodule TestUtils do
     |> String.trim()
   end
 
-  def assert_a11y(opts, filter) do
-    run_a11y(opts)
+  def assert_a11y(args, filter) do
+    run_a11y(args)
     |> A11yAudit.Assertions.assert_no_violations(filter: filter)
   end
 
-  def assert_a11y(opts) do
-    run_a11y(opts)
+  def assert_a11y(args) do
+    run_a11y(args)
     |> A11yAudit.Assertions.assert_no_violations()
   end
 


lib/dpul_collections_web/components/search_bar_component.ex
```diff
@@ -22,11 +22,7 @@ defmodule DpulCollectionsWeb.SearchBarComponent do
                 placeholder={gettext("Search")}
                 dir="auto"
               />
-              <button
-                id="search-button"
-                type="submit"
-                class="btn-secondary px-4 h-8 invisible flex-none"
-              >
+              <button id="search-button" type="submit" class="btn-secondary px-4 h-8 flex-none">
```
hackartisan, line 25: I would argue that the search button / search form is accessible the way it is, pending usability testing. It becomes visible when the input is focused. If there were a way to make the check pass without changing the behavior that would be nice. If not, we should consider whether this has implications for our design choices. And also probably remove the logic that controls its visibility, since that would be defunct.

hackartisan:
I made the background and accent colors lighter and darker, respectively, using the method outlined in https://www.learnui.design/blog/color-in-ui-design-a-practical-framework.html, to resolve color contrast. Here are some screenshots:

Screenshot 2025-07-16 at 12-04-46 Digital Collections
Screenshot 2025-07-16 at 12-07-44 Фрагмент з картини Волошки - Digital Collections
Screenshot 2025-07-16 at 12-08-01 Metadata - Фрагмент з картини Волошки - Digital Collections

tpendragon (approved):

PR #615, opened 2025-07-18
Add primary color background and corner cut decoration to call to action

### LINKED ISSUE #489: Browse callout iterations
### Summary or User Story

We may like to try some alternate layouts in the browse callout

### Acceptance Criteria

- [ ] Any member of the team who would like to propose alternate browse callout layouts has done so to their satisfaction, and the team has discussed those proposals

One or more PRs may address this issue. Work may or may not be merged as a result of this ticket.

hackartisan:
Also desaturates the mosaic images a bit, to forward the call to action
Also changes "browse" language to "explore" and removes "all items"

closes #489

drafted pending translations but posting for consideration

Screenshot 2025-07-18 at 17-09-32 Digital Collections

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -80,8 +80,8 @@
 }
 
 .browse-link {
-  padding-left: 50px;
-  clip-path: polygon(40px 0%, 100% 0%, 100% 100%, 0% 100%);
+  padding-left: 60px;
+  clip-path: polygon(50.72px 0%, 100% 0%, 100% 100%, 0% 100%);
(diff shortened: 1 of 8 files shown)


hackartisan:
I was able to force the angles to match by clipping to a 5-sided polygon; this requires making the search bar a little less tall in order to obscure the 5th side (if it's taller there's like a tiny knee-wall on the left before it starts to angle in)

Screenshot 2025-07-23 at 12 38 02 PM

Issue #611, opened 2025-07-18
Iterate on focused browse UI


tpendragon:
### Summary or User Story

We got a first attempt at a browse UI that allows recommending items in #600, but the design's not all the way there - it feels a little clunky, could use some juice, and feel a little better integrated. Let's see what we can do to make it feel better.

### Acceptance Criteria

- [ ] We're happy with the look, feel, and user interaction of the browse experience.

### First step

Try some things.

PR #629, opened 2025-07-22
Simplify browse similar items by removing the heart

hackartisan:
- Makes the focused item show as the first in the list
- fixes back button so you can get to all previous random sets

refs #611

Screenshot 2025-07-22 at 12 33 06 PM

### DIFF
--- a/lib/dpul_collections_web/live/browse_live.ex
+++ b/lib/dpul_collections_web/live/browse_live.ex
@@ -74,10 +74,10 @@ defmodule DpulCollectionsWeb.BrowseLive do
       <h1 id="browse-header" class="mb-2">{gettext("Browse")}</h1>
       <div class="mb-5 text-2xl w-full items-center">
         <div :if={!@focused_item} class="text-2xl mb-5">
-          Explore a random set of items from our collections.
+          {gettext("Exploring a random set of items from our collections.")}
         </div>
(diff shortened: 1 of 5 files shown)


lib/dpul_collections_web/live/browse_live.ex
```diff
@@ -59,55 +61,26 @@ defmodule DpulCollectionsWeb.BrowseLive do
 
   # If neither, generate a random seed and display random items.
   def handle_params(_params, _uri, socket) do
-    {:noreply, push_patch(socket, to: "/browse?r=#{Enum.random(1..1_000_000)}", replace: true)}
+    r = Enum.random(1..1_000_000)
+    socket = socket |> assign(r: r)
+    {:noreply, push_patch(socket, to: "/browse?r=#{r}")}
```
tpendragon, line 66: I don't think this is quite right yet. When you go to https://dpul-collections-staging.lib.princeton.edu/, click 'Browse', then click back you still get a new seed and now you can't go back to the home page. Maybe a new `handle_params` that does a replace: true if there's no `r` param, but there's an `r` in the socket, and it uses that `r`? And then still replace: true here.

tpendragon (changes requested):

tpendragon (approved):

PR #639, opened 2025-07-25
Configure solr from a json file

hackartisan:


hackartisan:
Try putting the file on the container as a volume, then reading it from there in code config:

https://github.com/pulibrary/princeton_ansible/blob/ffb693a8c138e8b928d5933ba223085fec64f7ca/nomad/traefik-wall/deploy/production.hcl#L62-L77

hackartisan:
close in favor of #691

PR #655, opened 2025-07-30
Aligns thumbnail with buttons at all screen sizes on item page

### LINKED ISSUE #646: item page thumbnail alignment
### Summary or User Story

There's a screen size at which the primary thumbnail is left aligned, or maybe a bit small. I think it should take up the full width, or maybe the width of the page should be constrained to fit it. At the least it should be centered.

Image

### Acceptance Criteria

- [ ] the primary thumbnail is aligned appropriately

### First step

css

sdellis:
Closes #646 

Screenshot at 618px wide is now aligned:

Screenshot 2025-07-30 at 12 23 25 PM

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -261,3 +261,9 @@
 .corner-cut {
   clip-path: polygon(0 50.72px,50.72px 0,100% 0,100% 50.72px,100% calc(100% - 50.72px),calc(100% - 50.72px) 100%,0 100%,0 calc(100% - 50.72px));
 }
+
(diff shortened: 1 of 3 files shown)


hackartisan (dismissed):

tpendragon (changes requested):
Unfortunately without that defined width it pushes the page down when the image loads in. You can replicate it by editing the HTML in your browser and giving it an invalid image URL - it shrinks down, that's what users see before the page loads in.

Issue #490, opened 2025-05-19
Conform to WCAG AA


hackartisan:
### Summary or User Story

Assess all pages of the site using browser plugins for adherence to WCAG AA accessibility requirements. Create and prioritize tickets that reflect work required to conform.

### Acceptance Criteria

- [ ] All accessibility violations are ticketed

### First step

run lighthouse or wave

PR #676, opened 2025-08-06
Work on some accessibility things

tpendragon:
- **Accessible labels for search.**
- **Add some semantic sections and aria-labels to improve VO**

Work towards #490

### DIFF
--- a/lib/dpul_collections_web/components/layouts/root.html.heex
+++ b/lib/dpul_collections_web/components/layouts/root.html.heex
@@ -12,12 +12,14 @@
     </script>
   </head>
   <body class="bg-white antialiased">
-    <a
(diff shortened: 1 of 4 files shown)


lib/dpul_collections_web/components/layouts/root.html.heex
```diff
@@ -12,12 +12,14 @@
     </script>
   </head>
   <body class="bg-white antialiased">
-    <a
-      class="transition left-0 bg-primary bg-gray-200 text-primary-content absolute p-3 m-3 -translate-y-16 focus:translate-y-0"
-      href="#main-content"
-    >
-      {gettext("Skip To Content")}
-    </a>
+    <nav id="skip-link" role="navigation" aria-label="Skip links">
```
hackartisan, line 15: translate

hackartisan (changes requested):
Nice, did you use a screen reader to find these issues?

just needs some more calls to gettext and run translation I think

Issue #675, opened 2025-08-06
Implement filters for Subject, Genre, Language, and Collection in search results


tpendragon:
### Summary or User Story

Implement #633 

### Full Screen Collapsed (Default)
![Image](https://github.com/user-attachments/assets/45a51d68-cfe8-4584-a363-fe501bce0109)

### Full Screen Expanded
![Image](https://github.com/user-attachments/assets/725ca4dd-2c0d-4b48-8a82-10e3d657f35e)

Clicking a facet will not change the results of facets in its tab. Adding another facet in a tab makes it an OR query.

Clicking a facet in another tab will update the other tabs to be restricted to the results of those selections.

All mockups can be adjusted in light of implementation requirements - treat is as a guideline rather than a pixel-for-pixel requirement. See comments below for mobile screens.

### Acceptance Criteria

- [x] desktop view UI + interactions are implemented roughly according to the mockup.
- [x] #849
- [x] #850
- [ ] Search filters (this may be useful for reference https://github.com/projectblacklight/blacklight/pull/3367)
- [ ] Pagination of the filter values
- [x] #848 
- [ ] For interactions that are best split out as sub-tickets, those tickets are made.

### First step

Start writing some tests.

PR #677, opened 2025-08-07
Make a bunch of fields facetable.

tpendragon:
These fields tokenize but also set docValues such that they can be
faceted upon. To do this migration I did a bunch of commands like

`find . -name "*.ex*" -type f -exec sed -i '' 's/holding_location_txtm/holding_location_txt_sort/g' {} +`

Work towards #675


hackartisan (commented):
This seems good, it looks like it will require reindexing when deployed, and also probably mess up dev environments so we should make sure everyone knows about it.

eliotjordan (approved):

PR #682, opened 2025-08-08
Updated related hydration cache entries

eliotjordan:
Works towards #75 

1. Updates hydration cache entries related to EphemeraTerms and FileSets.
2. Updates to parent in next chunk of work.


eliotjordan:
The process works except that the query to find related resources is so slow that Ecto times out. Even after converting related_data to jsonb and creating an index on it. I wonder if we're going to have to stash the IDs of related and ancestral resources directly on the HydrationCacheEntry.

Issue #667, opened 2025-08-04
Design UX for content warning


hackartisan:
### Summary or User Story

In other applications we embed the viewer from figgy and so we have to implement content warning logic from within the viewer. In this case we'll have access to content warning info throughout the application and we could have a different UI.

We've agreed we want to blur images throughout the application, until a user has clicked to show the images for that item.

We need design for the user interaction to show the images. Note that there will be content warnings of different lengths (although the ones we have for ephemera are short) and that the thumbnails are arranged in different ways affording different amounts of real estate.

Figgy adds a header and a link for more information to each content warning, see e.g. https://figgy.princeton.edu/catalog/d4292e58-25d7-4247-bf92-0a5e24ec75d1

### Acceptance Criteria

- [x] Design for viewer click-through is attached to this issue.
- [x] Design for search results click-through is attached to this issue.
- [x] Design for browse card click-through is attached to this issue.
- [x] Design for item page click-through is attached to this issue.
- [ ] Design for mobile click-through is attached to this issue.

### First step

drawings

sdellis:
Initial concept for Item Page. This approach could be replicated in Browse cards and Search rows. If there is agreement to this approach, I will create the mockups for the other pages.

<img width="1532" height="989" alt="Image" src="https://github.com/user-attachments/assets/21b6ecf8-dfba-4a53-842f-113633600dc4" />

<img width="1534" height="992" alt="Image" src="https://github.com/user-attachments/assets/1bf374ce-713d-45ed-af95-2c1aadb4bddc" />

hackartisan:
It looks nice! mockups for the other two would be helpful. I assume it should adjust the language of the link based on how many images there are.

I'm wondering if "sensitive content" is the right wording. Should it be "potentially harmful content"? Is there someone who should look over this little paragraph for us? Maybe it should just say, "Images are blurred because a content warning has been applied to this item. To view..."?

sdellis:
Good point, @hackartisan. I've updated that language in the Browse modal below. 

### Browse
Implementation Note: The "Why are the images blurry?" banner will cover the "Browse Similar Content" link until the user clicks "View Content". This should be implemented as a CSS overlay to avoid negotiating logic between the client and server. 

<img width="1249" height="810" alt="Image" src="https://github.com/user-attachments/assets/47f0945a-8bf5-4d94-94c2-b2a8f8ceabdd" />

[...]

Issue #371, opened 2025-03-20
Implement UX for content warning


hackartisan:
### Summary or User Story

Use designs developed in #667

In other applications we embed the viewer from figgy and so we have to implement content warning logic from within the viewer. In this case we'll have access to content warning info throughout the application and we could have a different UI.

- Do we need content warning UI in the search results, given the larger thumbnails?
- How should material with content warnings be presented in browse?
- How should the content warning work in an item page?

### Acceptance Criteria

- [x] Content warnings are indexed.
- [x] thumbnails on search, browse, and item pages are obfuscated if there is a content warning, and there's a click-through to show the images.
- [ ] recent items on home page are obfuscated if there is a content warning
- [ ] Viewers display a content warning that you click through if there is one.
- [x] images are not obfuscated again on another page or on page reload if the click-through has been accepted
- [ ] the click-through UX and design match the mock-ups created in #667 

## Implementation Notes

Try one obfuscation method and talk to the team.

### First step

Test for indexing content warnings.

hackartisan:
Here are the 2 figgy ephemera folders that have content warnings:
- d4292e58-25d7-4247-bf92-0a5e24ec75d1 
- eecc7710-1243-45e0-9efd-f61b63ad34ef

hackartisan:
for saving the click-through in session, these articles may be helpful:
- https://fly.io/phoenix-files/saving-and-restoring-liveview-state/
- https://fullstackphoenix.com/tutorials/set-session-values-from-liveview

PR #685, opened 2025-08-08
Obfuscate thumbnails throughout the site, store the ones that have been shown

hackartisan:
Blur images and provide an initial button (pending #667) to reveal them, saving the list of shown images in a cookie and loading it into server session to access as user navigates the site.

advances #371


hackartisan:
> I feel like there must be a way to de-duplicate some of the show_images passing around, but right now I really like this - those ways will probably become clearer. Good work!

Yeah, idk, I agree it is annoying! Object oriented programming, maybe? 😁

Issue #510, opened 2025-05-29
Design pass: buttons


hackartisan:
### Summary or User Story

Buttons need some standardization. Some things to consider:

- better visual distinction of a disabled button (change the color)
- all buttons should use all caps
- browse callout button needs more left / right padding
- are the arrows the same size for the left-pointing button and the right-pointing button?

### Acceptance Criteria

This section can be completed by the DLS team. Some requirements to consider
including:

- [ ] The UI implemented for this issue meets accessibility requirements
- [ ] New functionality is documented

### First step

This section can be completed by the DLS team

PR #706, opened 2025-08-18
Implements Button Designs

sdellis:
advances #510

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -156,7 +156,7 @@
     @apply enabled:border-0;
   }
   &:focus-visible {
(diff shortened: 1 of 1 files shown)


hackartisan:
I wonder if for the more button we should try a "chevron right" hero icon, either with a full-height vertical background or just on its own. I think one thing that makes it look small is that it has a horizontal orientation but needs to fit in a vertical space.

hackartisan (approved):
🎉

hackartisan:
I created https://github.com/pulibrary/dpul-collections/issues/749 so we can merge this and rebase other work

Issue #689, opened 2025-08-12
Investigate better use of vertical space in viewer pane


hackartisan:
### Summary or User Story

When you click to view an image, the amount of the website you still see, plus the thumbnails along the bottom of the screen, give the image very little space and it doesn't feel like you're able to view very much more detail. Investigate what we could do to give the image more space. Possibilities:

- Could / should the size of the "viewer" banner be adjusted?
- Could / should the viewer header be made smaller?
- Could the title bar of clover be removed and its functionality re-implemented in our title bar?
- Could the thumbnails go down the side when screens are wider than they are high? (we want to be mindful of mobile layouts as well)

### Acceptance Criteria

- [ ] straightforward solutions are ticketed for DLS to work
- [ ] changes that would need to be made to clover are ticketed in the clover project for discussion with clover developers

### First step

brainstorm

### Implementation Details

Mat had this to say about this and Clover:

> Hey Trey. Martin Uildriks from the Yiddish Book Center has a mixture of PRs in various that seek to make it easier for Clover to have more flexible layouts. One of interests is seeking to make Clover more modular like the Ramp viewer -- were subcomponents can be rendered within a layout. This direction makes the most sense to me.
> 
> As is --  you could hack at the CSS classes to make things a bit more friendly. And the Header can essentially be disabled using a handful of false values in your config. You then can render a custom header in your consuming application.

https://github.com/nulib/dc-nextjs/blob/deploy/staging/components/Clover/ViewerWrapper.tsx#L63-L78 is how they disable their header.

tpendragon:
Extracted https://github.com/pulibrary/dpul-collections/issues/733, closing this.

PR #716, opened 2025-08-21
Thumbnail on left

tpendragon:
Some iteration on #689 
Screen Shot 2025-08-21 at 2 37 18 PM

This puts thumbnails on the left side of the screen as long as you're above Tailwind's sm breakpoint.

### DIFF
--- a/assets/js/dpulc_viewer.jsx
+++ b/assets/js/dpulc_viewer.jsx
@@ -23,6 +23,7 @@ const handleCanvasIdCallback = (activeCanvasId, loadedCanvasIdx, pushEvent) => {
   }
 };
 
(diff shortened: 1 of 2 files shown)


assets/js/dpulc_viewer.jsx
```diff
@@ -1,20 +1,93 @@
...
+  // In vertical thumbnails, scroll to view.
+  scrollThumbnail(activeCanvasId, loadedCanvasIdx - 1)
   // Tell LiveView that we've changed the canvas so we can change the URL or
   // anything else, if necessary.
   if(typeof pushEvent === 'function') {
     pushEvent("changedCanvas", { "canvas_id": activeCanvasId })
   }
 };
+
+const scrollThumbnail = (activeCanvasId, loadedCanvasIdx) => {
```
hackartisan, line 26: How does this only impact vertical thumbnails? Is it the isOutsideVertically logic?

Can you add a comment to this function about why we need it for vertical thumbnails and not for horizontal? I assume there is similar code for the horizontal layout in clover itself?

tpendragon: @hackartisan Good idea! Added.

PR #718, opened 2025-08-22
Speed up tests by improving PhoenixTestPlaywright.

tpendragon:
This takes us from 164 to 118 seconds on my computer.

Related upstream PR: https://github.com/ftes/phoenix_test_playwright/pull/58

Work towards #679

PR #724, opened 2025-08-25
Use upstream phoenix_test_playwright improvements.

tpendragon:
Follow on to #718


test/dpul_collections_web/features/content_warnings_test.exs
```diff
@@ -41,7 +41,7 @@ defmodule DpulCollectionsWeb.Features.ContentWarningsTest do
       |> click_link("Why are the images blurred?")
       |> click_button("View content")
       |> refute_has("img.obfuscate")
```
hackartisan, line 43: Did you want a timeout on this refute?

hackartisan (changes requested):
I'm not sure I followed the entire upstream discussion. It sounded like the upstream project was going to have a default timeout of 0 or 1 on refute, relying on a convention that you first assert some state that would mean whatever you would need to wait for has happened, then the refute should pass. But based on these changes are we looking at a convention where we pass a .5 second timeout every time we use `refute_has`? And if so, can you clarify what happened upstream?

hackartisan (approved):
Okay this seems good, but now there's a new commit upstream, not sure if we want to adjust again before merging here. Approving but not merging just in case.

PR #773, opened 2025-09-17
Trying a couple small hydrator performance improvements

tpendragon:
I couldn't help myself looking at the Broadway dashboard and seeing resources get locked up in the batcher processor, the memory max out, and all the time get locked up querying for resources. I wanted to see if I could try this small improvement and see if it makes things better.

This prevents concurrent processors from waiting for records by always fetching at least 500 records in the producer, and prevents memory bloat by persisting records in `handle_message` to prevent sending large combined records to the batcher.

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/database_producer.ex
+++ b/lib/dpul_collections/indexing_pipeline/database_producer.ex
@@ -98,7 +98,7 @@ defmodule DpulCollections.IndexingPipeline.DatabaseProducer do
   end
 
   defp calculate_stored_demand(total_demand, fulfilled_demand)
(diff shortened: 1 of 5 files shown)


tpendragon:
The dashboard used to show these at random numbers all the time, showing that they were rarely processing at capacity. Now it looks like this:

Screen Shot 2025-09-17 at 3 25 05 PM


Also it's processing skips at ~ 3200/second vs 2100/second.

tpendragon:
With this hydration takes 1 hour 46 minutes, before that the fastest we've ever done it was in 3 hours.

Issue #755, opened 2025-09-11
SAE has a collection landing page


tpendragon:
### Summary or User Story

Our first slice is to create a basic collection landing page for South Asian Ephemera

The goal of collection landing pages is to allow people who visit to quickly evaluate if that collection is interesting to them and then draw them further into its items.

### Acceptance Criteria

- [x] SAE is indexed with the slug, title, and description.
    - [x] https://github.com/pulibrary/figgy/issues/6801
    - [x] https://github.com/pulibrary/dpul-collections/issues/756
- [ ] There's a page at `https://digital-collections.princeton.edu/collections/sae` which has:
    - [x] https://github.com/pulibrary/dpul-collections/issues/757

### First step

Go down the list of tickets.

PR #775, opened 2025-09-18
Real data for Collection Page

### LINKED ISSUE #756: Index Ephemera Project for SAE
### Summary or User Story

To support #755, SAE should be indexed so we have a place to pull the title, tagline, and description.

### Acceptance Criteria

- [ ] SAE has a solr document that we can use to build the page in #755 
- [ ] Search results continue to work

### First step

Write a test.

tpendragon:
Work towards #755 
closes #756 

This does:

1. Indexes published Ephemera Projects
2. Indexes Featurable.
3. Uses indexed data to populate collection pages.
4. Provides a -very- basic search result view (We'll likely want to do something totally different here to differentiate them from items, but it's a new feature we might be able to get some buy-in with.)
5. Increments the cache version in staging.
6. Fixes deploys so that it doesn't accidentally pull a cached version of artifacts by pulling from a sha instead of a branch reference.

This is deployed now here: https://dpul-collections-staging.lib.princeton.edu/collections/sae

Screen Shot 2025-09-24 at 10 21 42 AM
Screen Shot 2025-09-24 at 10 21 46 AM
Screen Shot 2025-09-24 at 10 22 37 AM


assets/js/hooks.js
```diff
@@ -194,9 +189,8 @@ Hooks.ResponsivePills = {
...
     // Reserve space for the more button
     // The more button is invisible rather than hidden, so we can get its
-    // height/width. The first pill doesn't have a gap on its left, so we don't
-    // add a gap here for calculating.
-    const moreButtonWidth = moreButton.offsetWidth
+    // height/width.
+    const moreButtonWidth = this.moreButton.offsetWidth + gap
 
     // Calculate how many items fit on one line
     for (let i = 0; i < allPillItems.length; i++) {
```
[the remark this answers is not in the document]
tpendragon: It has to hide every pill except the ones that aren't on the first line, to do that it has to see how many fit on the first line and leave room for the more button.

tpendragon:
> The "more" and "less" buttons seem to be missing from the mobile view.

That's interesting, that looks to me like the JS didn't work.

lib/dpul_collections/solr.ex
```diff
@@ -14,12 +15,53 @@ defmodule DpulCollections.Solr do
...
+    |> Map.new()
+  end
+
   @query_field_list [
     "id",
     "title_ss",
     "display_date_s",
     "file_count_i",
     "detectlang_ss",
+    "authoritative_slug_s",
```
hackartisan, line 63: This is a little confusing, if there's an authoritative slug then what does that mean about the other slug? Why can't we use the other slug field for collection slugs, given that we have the resource type?

test/dpul_collections_web/live/browse_live_test.exs
```diff
@@ -14,6 +14,15 @@ defmodule DpulCollectionsWeb.BrowseLiveTest do
     assert redirected_to(conn, 302) =~ "/browse?r="
   end
 
+  test "browsing when there's a collection", %{conn: conn} do
```
hackartisan, line 17: Can you make the title of this test say what will happen?

test/dpul_collections_web/live/search_live_test.exs
```diff
@@ -42,6 +42,17 @@ defmodule DpulCollectionsWeb.SearchLiveTest do
            |> Enum.any?()
   end
 
+  test "GET /search with an ephemera project title", %{conn: conn} do
```
hackartisan, line 45: "will include the project in results"?

test/support/figgy_test_support.ex
```diff
@@ -99,6 +99,14 @@ defmodule FiggyTestSupport do
     solr_doc
   end
 
+  def index_record_id_directly(id) do
```
hackartisan, line 102: yesss

hackartisan (changes requested):
this looks really good, i have just a few small comments

lib/dpul_collections/solr.ex
tpendragon: Oh yes I can probably refactor this out now. The reason I had it was because I was using Item structs in the search results view and things were gettin' weird for some reason, now that I use the collection structs I think we can do better. Thank you

tpendragon: OH, wait, no, the reason is because Solr's generating `slug_s` and it overrode whatever I put there.

hackartisan (approved):
👍🏻

PR #784, opened 2025-09-22
Fix searching and make stemmed searches for controlled vocabs work.

### LINKED ISSUE #783: Searching is only searching description/title
### Summary or User Story

When we changed the fields to `txt_sort` I missed the `qf` in the query parser. I also noticed that we're not stemming subjects, so "Art" doesn't match "Arts" in the subject field.

### Acceptance Criteria

- [ ] Searching all the metadata fields we picked before works again.
- [ ] txt_sort fields stem as english, as all of our controlled vocabs are in english.

### First step

Write some tests.

tpendragon:
Closes #783


tpendragon:
With this, "Art" in staging returns 7,098 results, in prod it returns 178.

PR #805, opened 2025-09-26
Fix race condition caused by non-serialized writes to markers

tpendragon:
1. Adds a test.
    * It turns out the problem wasn't the transaction, it was that the Sandbox adapter uses a single database connection for the whole test. To get a race condition to happen, we have to check out multiple connections - one per processor/batch processor. We can do that similar to the Broadway documentation for Ecto, using telemetry to explicitly check out a fresh connection.
2. Adds an advisory lock per cache version to serialize writes


tpendragon:
This does fix the problem, but the full pipeline test is still unreliable.

eliotjordan (approved):

PR #819, opened 2025-10-03
Add Project tagline to Item Page

### LINKED ISSUE #758: Collections with taglines should display them in the item view
### Summary or User Story

Taglines are short messages meant to draw folks to collections they might be interested in. In the item view we currently display the title - we should also display that short tagline. This may require some design to make it look alright.

### Acceptance Criteria

- [ ] Item views have taglines for the project they're a part of.

### First step

Index tag line as part of an item.

sdellis:
Closes #758

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -20,7 +20,6 @@
   --color-sage: #d0d3ac;
   --color-dark-sage: #9fa06c;
   --color-cloud: #c1ccc3;
-  --color-dark-cloud: #b0bab2f;
   --color-rust: #a0562c;
   --color-dark-blue: #3a4151;
   /* Button Text on Dark Blue */


assets/css/app.css
```diff
@@ -20,6 +20,7 @@
   --color-sage: #d0d3ac;
   --color-dark-sage: #9fa06c;
   --color-cloud: #c1ccc3;
+  --color-dark-cloud: #b0bab2f;
```
tpendragon, line 23: Do we have a semantic name for this color?

lib/dpul_collections/item.ex
```diff
@@ -100,6 +101,7 @@ defmodule DpulCollections.Item do
       {gettext("Institutional Information"),
        [
          {:project, gettext("Ephemera Project")},
+         {:tagline, gettext("Tagline")},
```
tpendragon, line 104: Yay, nice catch!

tpendragon (changes requested):
This seems nice - I just wonder if that color should have a semantic color so we don't lose track of when to use it.

Looking at the screenshot I was worried about the contrast of that collections header, but knew the [automated accessibility tests](https://github.com/pulibrary/dpul-collections/blob/246d03722eccd7b98bbb261d625e8d2b0a5480ac/test/dpul_collections_web/features/item_test.exs#L170-L174) should catch it - but then realized that the feature test for that uses mock solr documents, which won't have this new field:

https://github.com/pulibrary/dpul-collections/blob/246d03722eccd7b98bbb261d625e8d2b0a5480ac/test/dpul_collections_web/features/item_test.exs#L10

https://github.com/pulibrary/dpul-collections/blob/246d03722eccd7b98bbb261d625e8d2b0a5480ac/test/support/solr_test_support.ex#L40-L67

so it wouldn't get tested. I sent a PR against this branch that should make sure it's tested, if that looks good to you: https://github.com/pulibrary/dpul-collections/pull/833

Turns out the contrast was fine, but it'd be good to keep testing it I think? Curious what you think.

Issue #809, opened 2025-09-29
Implement User Sets


sdellis:
### Summary or User Story
Implement User Sets according to the screen designs provided in this Epic ticket. 

user stories in #741

Note: Private Sets are not part of this ticket.

### Acceptance Criteria

- [x] The UI implemented for this issue meets accessibility requirements.
- [x] User authentication feature is implemented with "magic links," without the need for a password.
- [x] Users can save items into personal sets from the Browse and Search Results screens
- [x] #896
- [x] Users can view all their own sets on a "my sets" page
- [x] #899
- [x] Users can delete items from their own sets. (you can do this right now by clicking save and then removing it from the set)
- [x] #897
- [x] #898
   - [x] My Account and Language main menu options are collapsed into a single dropdown on mobile screens.
- [x] #930
- [x] put sets in the menu
- [x] Make sure calls to Layouts.app pass current scope
- [ ] Turn the feature on in prod

### Implementation notes

For implementing other pages that depend on the existence of a logged in user see https://hexdocs.pm/phoenix/scopes.html

### Search Results with Save to Set
![Image](https://github.com/user-attachments/assets/980e9b53-c6b6-4247-8060-c90ac92889f1)

### Item Page with Save to Set
![Image](https://github.com/user-attachments/assets/05732055-9858-4503-b45f-ebb5481052a3)

### Browse Page with Save to Set
![Image](https://github.com/user-attachments/assets/bde0eb1a-5c1b-4843-b874-ae5926026b76)

### User Login
![Image](https://github.com/user-attachments/assets/cf65d0a4-fa38-4060-966e-61f84b889257)

### User Verify
![Image](https://github.com/user-attachments/assets/2bbd18df-3f87-4f28-a971-89afd9c98778)

### Manage User Sets
![Image](https://github.com/user-attachments/assets/0d341106-ed41-49b2-a664-0b8e7c28b261)

### User Personal Set (Logged in)
![Image](https://github.com/user-attachments/assets/ecf52a0b-f326-4d0f-a6b1-71140b3073de)

### User Set - (Not Logged in / Shared)
![Image](https://github.com/user-attachments/assets/b6a31b14-149e-4a2e-9f1d-7d3548ba9139)

### Save to Set (Modal)
![Image](https://github.com/user-attachments/assets/247b2752-e6fb-482a-bfc6-9981b89da8c9)

### Create/Edit Set (Modal)
[...]

PR #864, opened 2025-10-16
User Sets: Add Modal & Models

tpendragon:
Work towards #809 

## What this does

* Adds `UserSets.Set` and `UserSets.SetItem`, and links Sets to Users.
* Adds a LiveComponent for `AddToSet` which controls the modal that pops up and lets users add modals.
* Adds the button to initiate the modal to both the browse and search results.
* Modal lets you create sets, add an item to a set, remove items from sets, and shows which of your sets an item is already in.
* Locks all of the above behind our user registration feature flag.
* Adds helpers and infrastructure so Playwright tests can log a user in.
* Adds the ability for the server to trigger a modal to open.

## Demo

https://github.com/user-attachments/assets/b8425ee2-9f29-4052-9f1d-c5207a5bcdb9

## Notes

A few things that were challenging or surprising in here:

- Feature tests while logging in meant we had to figure out the SQL sandbox, but now that's done and we have a way to log in for those. Mostly pulled that from https://hexdocs.pm/phoenix_test_playwright/PhoenixTest.Playwright.html#module-ecto-sql-sandbox and https://hexdocs.pm/phoenix_ecto/Phoenix.Ecto.SQL.Sandbox.html#module-acceptance-tests-with-liveviews and https://hexdocs.pm/phoenix_test_playwright/PhoenixTest.Playwright.html#add_session_cookie/3
- Had to figure out nested forms.
- The generators create a bunch of methods, and it's unclear which of those we need quite yet. There'll be some cleanup for sure.
- To get current_scope in everything I had to move the router around.
- The modal kept removing itself when it was updated dynamically, because the JS adds the `open` attribute and then Phoenix overwrote it - had to add `phx-mounted={JS.ignore_attributes("open")}` - thanks discord!
- I needed a way to open the modal only after its content was updated, otherwise it flashed weirdly - to do that I needed an event that the server could fire that told the Modal to open.
- Adding a query to efficiently get all the user sets for a user along with how many items that thing has, and if it has the item I care about, was pretty tough. Complicated SQL -> Ecto - thanks again discord.

Things I've learned so far:

- The scope thing for users is pretty cool!
- `to_form` for ChangeSets is pretty neat.
- A "LiveComponent" should hopefully let us put this button on every page without duplicating event handling.

### DIFF
--- a/lib/dpul_collections_web/components/browse_item.ex
+++ b/lib/dpul_collections_web/components/browse_item.ex
@@ -79,17 +79,15 @@ defmodule DpulCollectionsWeb.BrowseItem do
         @class
       ]}
     >
-      <div class="absolute p-4 inset-0 w-full">
-        <div class="w-full flex items-top justify-end gap-2">
-          <.card_button
-            :if={@likeable?}
-            patch={~p"/browse/focus/#{@item.id}"}
-            phx-click={JS.dispatch("dpulc:scrollTop")}
-            icon="iconoir:binocular"
-            label={gettext("Similar")}
-          />
-          <UserSets.AddToSetComponent.add_button :if={@current_scope} item_id={@item.id} />
-        </div>
+      <div class="absolute p-4 right-0 flex gap-2">
+        <.card_button
+          :if={@likeable?}
+          patch={~p"/browse/focus/#{@item.id}"}
+          phx-click={JS.dispatch("dpulc:scrollTop")}
+          icon="iconoir:binocular"
+          label={gettext("Similar")}
+        />
+        <UserSets.AddToSetComponent.add_button :if={@current_scope} item_id={@item.id} />
       </div>
       <div
         :if={Helpers.obfuscate_item?(assigns)}


test/support/live_acceptance.ex
```diff
@@ -0,0 +1,15 @@
+defmodule LiveAcceptance do
```
hackartisan: module docs?

lib/dpul_collections_web/live/user_sets/add_to_set_component.ex
```diff
@@ -17,17 +25,105 @@ defmodule DpulCollectionsWeb.UserSets.AddToSetComponent do
...
+      <.input type="text" required={true} label="Set name" field={@set_form[:title]} />
+      <.input type="textarea" label="Set description" field={@set_form[:description]} />
+      <.primary_button>
+        Create Set
+      </.primary_button>
+    </.form>
+    """
+  end
+
+  def append(assigns) do
```
hackartisan: maybe call it `select_set`?

lib/dpul_collections/user_sets.ex
```diff
@@ -44,6 +45,40 @@ defmodule DpulCollections.UserSets do
     Repo.all_by(Set, user_id: scope.user.id)
   end
 
+  def list_user_sets_for_addition(%Scope{} = scope, solr_id \\ nil) do
```
hackartisan: Can we add spec to this function? It's not clear until you read the test what it's returning, in particular what "has_solr_id" means.

lib/dpul_collections_web/components/core_components.ex
```diff
@@ -414,6 +414,7 @@ defmodule DpulCollectionsWeb.CoreComponents do
       id={@id}
       open={assigns[:open]}
       phx-hook="Dialog"
+      phx-mounted={JS.ignore_attributes("open")}
```
hackartisan: maybe a code comment here?

hackartisan (changes requested):

Issue #521, opened 2025-06-03
Raise local exception for non-200 solr responses


hackartisan:
### Summary or User Story

Here's an example of a 404 returned for a solr query: https://app.honeybadger.io/projects/130995/faults/119092242

Our code should handle this by raising a local exception that tells us that solr is down.

This will improve the readability of honeybadger errors.

### Acceptance Criteria

- [ ] the dpul-c solr module raises exceptions when the response from solr is anything other than a 200

### First step

look at the honeybadger error and write a test mocking that response.

PR #866, opened 2025-10-17
Extract a Solr client.

tpendragon:
Pulling all the Solr interactions out to a client should make it easier to work on things like #521 and standardize our return values.


lib/dpul_collections/solr.ex
```diff
@@ -322,11 +323,8 @@ defmodule DpulCollections.Solr do
   def add(docs, index \\ Index.read_index())
 
   def add(docs, index) when is_list(docs) do
```
hackartisan: these both have the same implementation now? should we just have one of them?

tpendragon: Yeah, I was kinda going with like...if I wrote this as a library, what would it be. Might be premature 🐣

hackartisan (approved):
ty!

PR #869, opened 2025-10-21
Remove unhelpful alt text

### LINKED ISSUE #824: Remove alt text from search results, cards, and item view images
### Summary or User Story

Our images don't have useful alt text right now, and "image 1" just sounds bad. All of our images from the repository should have `alt=""` - none is better than bad, when it comes to reading the page. Hopefully in many cases there's enough information on the page, and if there's not then we need a feature for creating alt for images anyways.

### Acceptance Criteria

- [ ] Images without useful alt descriptions have an empty alt tag.

### First step

Search for where we're generating alt tags like "Image 1" and remove them.

eliotjordan:
Closes #824

### DIFF
--- a/lib/dpul_collections_web/components/browse_item.ex
+++ b/lib/dpul_collections_web/components/browse_item.ex
@@ -205,7 +205,7 @@ defmodule DpulCollectionsWeb.BrowseItem do
         "thumbnail-#{@item.id}"
       ]}
       src={thumbnail_url(assigns)}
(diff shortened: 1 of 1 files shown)


hackartisan (changes requested):
I tried it with the screen reader on the search page, it makes voiceover skip those images 👍🏻 

There's alt text on the browse cards, too, let's do the same thing there

PR #872, opened 2025-10-22
Generate an accurate file count

### LINKED ISSUE #810: Search results say "3 images" but only show 2
### Summary or User Story

Example item with this issue: https://figgy.princeton.edu/concern/ephemera_folders/6d2be738-626f-4b3d-a17a-655501d2ad9e/file_manager

screenshot:
Image

It looks like there are items with different types of files that aren't going to display in dpul-c, but we're not filtering them all out. Should we have an allow list for file types?

### Acceptance Criteria

- [ ] Items have an accurate file count
- [ ] Check PDFs

### First step

Figure out why that record's showing 3 images, and find a reliable way to reliably prevent that.

eliotjordan:
Closes #810

- **Add non-image file fixture**
- **Generate an accurate file count**

Screenshot 2025-10-22 at 1 18 04 PM

**Versus:**
Screenshot 2025-10-22 at 1 37 45 PM


hackartisan (approved):
👍🏻

Issue #849, opened 2025-10-13
Mobile view for filters in search results


tpendragon:
### Summary or User Story

Part of #675 

### Acceptance Criteria

- [ ] There's a mobile view for filters that reflects the design in #675 

### First step

Write a test.

PR #874, opened 2025-10-23
Extract viewer and metadata pane into a live component

hackartisan:
This was done in pursuit of #849 but ended up not being relevant there. Do we want it anyway? it might improve readability of the item page.


eliotjordan (approved):
This seems like a good idea. 👍

PR #895, opened 2025-10-31
Add sets page

hackartisan:
Incremental - not feature complete

- **Remove unused href prop**
- **Add sets page**

Advances #809

screenshot: 
Screenshot 2025-10-31 at 11-37-50


tpendragon (approved):
Just a small comment, but non-blocking. The card is a little odd still, and goes to the item page for the first item, but it's a great step forward.

PR #918, opened 2025-11-06
Add a user set page

### LINKED ISSUE #897: Anyone with the link can view an individual set and all its items at `/sets/:id`
See https://github.com/pulibrary/dpul-collections/issues/809 for details and mockups

tpendragon:
Closes #897

Own set:

Screen Shot 2025-11-06 at 3 11 21 PM

Anonymous:

Screen Shot 2025-11-06 at 3 11 44 PM

### DIFF
--- a/lib/dpul_collections_web/live/user_sets_live/show.ex
+++ b/lib/dpul_collections_web/live/user_sets_live/show.ex
@@ -66,7 +66,7 @@ defmodule DpulCollectionsWeb.UserSetsLive.Show do
         </div>
         <hr />
         <h2>{length(@user_set.set_items)} Items</h2>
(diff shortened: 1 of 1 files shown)


eliotjordan (commented):
A couple of comments:

1. I wonder if the "Save"  buttons on each card make sense here. It's more like "Delete or Move to Another set." I don't know how to say thay succinctly, though.
2. The cards don't flow on a smaller screen size/mobile like they do on the Explore page. Is that something to tackle here on in another ticket?

Otherwise, this looks great!

tpendragon:
@eliotjordan Thanks for catching the flow, I can fix that here tomorrow morning.

I left save in there so that other users can save those items.

Issue #950, opened 2025-11-14
Iteration: Collection page


hackartisan:
### Summary or User Story

Some comments in #934 warrant further discussion and potential iteration, including
- horizontal margin difference between top section and "learn more" section
- ways to address differences in hero image crop area between different screen sizes
- is there a way to make the facet summaries more prominent?

Ideas to try:

- [x] Make the margins match at all the screen levels (Trey)
- [x] Make the tagline take up 60% of the available width in the top
- [x] Flip the filters to the left in the learn more area, keep them 1/3 of the width.
- [x] Move the filters under the learn more heading, only have one column in that area.
- [x] Keep the landscape orientation on the thumbnail, but never let its aspect ratio exceed 4:3 presumably by making the text take up more space, or by growing the image height. (Shaun gonna try this one)
- [x] Constrain the hero area to be 4/3 - it has a max width.
- [x] Write another ticket to try out ways to add more featured items prominently in the hero area.

### First step

Meet as a team

hackartisan:
Flipping the 60 / 40 areas looked like this (we didn't like it):

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/70687a53-9770-4749-bef3-01b1ccde2d93" />

hackartisan:
closed by #955 and #958

PR #955, opened 2025-11-14
makes margins responsive

sdellis:
ref #950 

### Large Screen
Screenshot 2025-11-14 at 3 32 06 PM

### Laptop
Screenshot 2025-11-14 at 3 32 49 PM

### Mobile
Screenshot 2025-11-14 at 3 33 06 PM


hackartisan:
> @hackartisan it does something different and is harder to read at large screen widths. That adds more margins as the screen gets bigger so that the single column text doesn't spread too wide.

😕 it looks the same to me when I try it. And it fixes a weird cut-off bug at a certain screen size (see screenshot)

Screenshot 2025-11-14 at 4 21 48 PM

hackartisan (approved):

Issue #975, opened 2025-11-19
Share modal is squished on modal


tpendragon:
### Summary or User Story

<img width="523" height="631" alt="Image" src="https://github.com/user-attachments/assets/5d10bd9d-f0a2-4e76-84dd-1708e73160d1" />

The share modal shouldn't be squished.

### Acceptance Criteria

- [ ] Share modal expands to fill the screen's available space.

### First step

Play with the CSS.

PR #995, opened 2025-11-21
Fixes overflow issue on mobile share modal

sdellis:
Ref #975 

This PR has three commits. The first fixes the mobile UI issue. The second fixes a failing test that would prevent a merge. (The third is just formatting.)


sdellis:
### Mobile View 
Note: The share modal on larger screens is unchanged. 

Screenshot 2025-11-21 at 4 40 04 PM

hackartisan (changes requested):
This looks better! I just have a couple of small suggestions for consideration.

PR #1074, opened 2026-04-03
Add a SolrDocument

eliotjordan:
Refactor:

- Extracts logic for generating solr documents from HydrationCache entries into a single module.
- Most of the logic was moved from CombinedFiggyResource.
- Simplifies HydrationCacheEntry so it just an Ecto Schema - logic for building an enriched "combined" figgy resource is moved elsewhere
- Consolidates and simplifies the process of building solr documents so it is less surprising and easier to reason about.


test/dpul_collections/item_test.exs
```diff
@@ -19,10 +17,7 @@ defmodule DpulCollections.ItemTest do
     end
 
     test "can build from an Ephemera Project" do
```
tpendragon: This is the main thing we lose I think. I've got this dream that the indexing pipeline could just filter at each step, and the specifics of hydration and implementation is a purely functional problem we can iterate on and test outside of the scope of the hand-off between stages, and that might be easier to understand.

Basically that "creating a solr document from a Figgy resource", "running a figgy resource through an effective indexing pipeline", and "choosing which things should be part of the site" can be three different and easily identified responsibilities. The fact that you made this refactor may mean that's not true though, and that separation is too hard to hold together.

lib/dpul_collections/indexing_pipeline/figgy/solr_document.ex
```diff
@@ -0,0 +1,509 @@
+defmodule DpulCollections.IndexingPipeline.Figgy.SolrDocument do
```
tpendragon: Having a SolrDocument as its own module is great! Heck, maybe we might want to turn it into a struct at some point, as an in-code way to formalize our solr data structure?

tpendragon (dismissed):
I think this is a great refactor that we should talk about! Thanks for undertaking it. The last refactor here (https://github.com/pulibrary/dpul-collections/pull/746) said it was "a successful refactor if it's easier to understand and guess where modifications to the pipeline happen and what's going on at each step" - some of this reverts what happened there, so maybe that's a sign that it wasn't successful.

hackartisan (commented):
I'm not sure. I think I don't have a good handle on the problem. I do sort of like having all the solr index field names in a solr-y place.

Here are some notes from the last time we talked about this `to_` vs `from_` stuff https://docs.google.com/document/d/1hHq82XqYs29iSFnJMTvPKoIBMgk6YQM-95w_vTRvGrM/edit?tab=t.0#heading=h.efa1183x2503

It seems like we should talk.

hackartisan (commented):
One thing that's nice about this is that you can much more easily find the test to use when you need to index a new value. Some other kind of reorganization might also be effective at this, like moving some of the tests in hydration_cache_entry_test to combined_figgy_resource_test

tpendragon (commented):
Alright I rebased and adjusted it a little so it does a little less. I didn't adjust it as much as I expected - I left `from_` for instance, because I think it's a step towards the next refactors and because it turned out to be weirdly hard to not do that.

lib/dpul_collections/indexing_pipeline/figgy/combined_figgy_resource.ex
```diff
@@ -1,712 +1,12 @@
 defmodule DpulCollections.IndexingPipeline.Figgy.CombinedFiggyResource do
-  alias DpulCollections.IndexingPipeline.Figgy
-  require Logger
+  @moduledoc """
```
hackartisan, line 2: 💯

hackartisan (approved):
this seems like a reasonable step

PR #1125, opened 2026-04-16
Proposal: Adjust home page filters to have manuscripts & digital collections

tpendragon:
Folks keep asking how to get to the list of collections, and now that we have manuscripts in there we should advertise them. We removed "manuscripts" in the past because we didn't have any in Ephemera.

Screenshot 2026-04-16 at 11 18 49 AM

### DIFF
--- a/lib/dpul_collections_web/live/home_live.ex
+++ b/lib/dpul_collections_web/live/home_live.ex
@@ -42,11 +42,11 @@ defmodule DpulCollectionsWeb.HomeLive do
                 )}
               </div>
               <div class="flex flex-wrap justify-center text-dark-text gap-2">
+                <.format_link filter="digital collection" label={gettext("Digital Collections")} />
                 <.format_link filter="posters" label={gettext("Posters")} />
                 <.format_link filter="pamphlets" label={gettext("Pamphlets")} />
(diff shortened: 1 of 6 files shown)


eliotjordan:
I'm not sure this is particularly user friendly. The button is called "digital collections" which is the name of the web site! I don't think it's obvious that clicking that will take a user to a list of "Collections". I think it's because we are conflating a traditional filter (type = poster) with something more general. Narrowing scope vs. showing which buckets all of our records are a part of.

Although it's more work than a button, a new row, something like what northwestern does, makes much more sense to me: https://dc.library.northwestern.edu/

image

tpendragon:
@eliotjordan I agree we need something better, this was just a quick stop-gap. I think I'm hearing your opinion is "it's better to have nothing there than this, given how confusing it is"?

hackartisan:
What about a button to the left of "explore" that says "browse all collections"

similar to the wording of "browse collection" that's on the collection page

eliotjordan:
I think a "browse all collections" button is better, but.... I was hoping we could add a "browse" button next to "explore" that takes the user to the search page so they can use filters. Three similar buttons there is a little much. I think this is an OK temporary measure to get that functionality in for our PO, but I would really like to build out something more robust later.

PR #1126, opened 2026-04-16
Reduce language-specific query fields

### LINKED ISSUE #1082: Research: Long queries are erroring
### Summary or User Story

Any long query sent to Solr is sending a maxClauses error. Maybe we're sending them wrong?

https://app.honeybadger.io/projects/130995/faults/128469112/01KN7FBJQ3HK4RB896H9ZQ2NZV?q=-context.maintenance_window%3A%22true%22+occurred.after%3A%228+days+ago%22

### Acceptance Criteria

- [ ] Research some options for fixing this problem and report back
- [ ] The team picks an option.

### A few options we brainstormed

1. Lower the number of qf fields by doing less languages
2. Don't query every qf field, do language detection at query time and pick them dynamically.
    * If someone sends a query that's in spanish, detect that and set the "qf" to be "title_es".
    * Detect languages with something like https://github.com/joshrotenberg/lingua_ex
    * People can select the language they're searching in to try to get better results.
    * Send multiple queries for every submission
    * Use the user's browser language
3. Don't do this language business until we have real queries that we're optimizing for.
4. See if there's a way to make Solr handle this.
5. Put everything in a big `keyword` field, just search that.
6. People could select the language they're searching in to try to get better results.

Keep in mind #1107 - we'll have a lot more fields we'll want to add, so whatever our solution is will have to scale.

### First step

Reproduce in test.

### LINKED ISSUE #1127: Use `langid.allowlist` instead of `langid.whitelist` in Solr
### Summary or User Story

Fits with our [inclusive language](https://github.com/pulibrary/pul-it-handbook/blob/main/norms/inclusive_language.md) policy and it's now supported.

### Acceptance Criteria

- [ ] We use `langid.allowlist`

### First step

Update it.

eliotjordan:
Closes #1082
Closes #1127

- **Reduce language-specific query fields**
- **Use langid.allowlist**

### DIFF
--- a/solr/conf/solrconfig.xml
+++ b/solr/conf/solrconfig.xml
@@ -699,75 +699,19 @@
 
         title_txtm
         title_txtm_nolang
-        title_txtm_en
         title_txtm_ar
-        title_txtm_bg
-        title_txtm_ca
-        title_txtm_cjk
-        title_txtm_cz
-        title_txtm_da
-        title_txtm_de
-        title_txtm_el
+        title_txtm_en
         title_txtm_es
-        title_txtm_et
-        title_txtm_eu
-        title_txtm_fa
-        title_txtm_fi
-        title_txtm_fr
-        title_txtm_ga
-        title_txtm_gl
-        title_txtm_hi
-        title_txtm_hu
-        title_txtm_hy
-        title_txtm_id
-        title_txtm_it
-        title_txtm_ja
-        title_txtm_ko
-        title_txtm_lv
-        title_txtm_nl
-        title_txtm_no
         title_txtm_pt
-        title_txtm_ro
         title_txtm_ru
-        title_txtm_sv
-        title_txtm_th
-        title_txtm_tr
 
         summary_txtm
         summary_txtm_nolang
-        summary_txtm_en
         summary_txtm_ar
-        summary_txtm_bg
-        summary_txtm_ca
-        summary_txtm_cjk
-        summary_txtm_cz
-        summary_txtm_da
-        summary_txtm_de
-        summary_txtm_el
+        summary_txtm_en
         summary_txtm_es
-        summary_txtm_et
-        summary_txtm_eu
-        summary_txtm_fa
-        summary_txtm_fi
-        summary_txtm_fr
-        summary_txtm_ga
-        summary_txtm_gl
-        summary_txtm_hi
-        summary_txtm_hu
-        summary_txtm_hy
-        summary_txtm_id
-        summary_txtm_it
-        summary_txtm_ja
-        summary_txtm_ko
-        summary_txtm_lv
-        summary_txtm_nl
-        summary_txtm_no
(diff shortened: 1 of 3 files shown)


solr/conf/solrconfig.xml
```diff
@@ -1181,7 +1125,7 @@
     <str name="langid.map.keepOrig">true</str>
     <str name="langid.map.individual">true</str>
     <str name="langid.threshold">0.8</str>
-    <str name="langid.whitelist">en,ar,bg,ca,cj,cz,da,de,el,es,et,eu,fa,fi,fr,ga,gl,hi,hu,hy,id,it,ja,ko,lv,nl,no,pt,ro,ru,sv,th,tr,zh,zh-cn,zh-tw</str>
+    <str name="langid.allowlist">en,ar,bg,ca,cj,cz,da,de,el,es,et,eu,fa,fi,fr,ga,gl,hi,hu,hy,id,it,ja,ko,lv,nl,no,pt,ro,ru,sv,th,tr,zh,zh-cn,zh-tw</str>
```
tpendragon, line 1128: Should this be a smaller list, so it doesn't classify into fields we don't query for?

tpendragon (changes requested):

solr/conf/solrconfig.xml
eliotjordan: Personally, I'm in favor of leaving them. This allows us to make use of the additional language fields in the future if we need them. For example, if we decide to do client-side language detection, we can iterate on that without a full re-index. Or if we ingest new content and discover that CJK records way outnumber Russian language records we can easily make a switch to prioritize those.

tpendragon: What I'm worried about is that it will index into `title_txtm_zh` and then you can't search the title

eliotjordan: We have [langid.map.keepOrig set to true](https://github.com/pulibrary/dpul-collections/blob/main/solr/conf/solrconfig.xml#L1181), so it should still be searchable because the original content stays in title_txtm (which is in qf).

tpendragon: Alright, sounds good.

tpendragon (approved):
I'm happy with this, but eager to hear from others too. Thank you for this research.

It'll need a `pul_solr` PR.

hackartisan:
This seems like a reasonable compromise. I sort of wonder how many documents are losing this feature because they weren't in the top 5 languages. But I wouldn't block over it. I like the new tests.

eliotjordan:
@hackartisan I did some counting based on our language facets.
1. With this change, 2,032 records no longer have the enhanced qf field processing that used to.
2. Overall, there are 8,182 records that don't have this enhanced processing.
3. The 5 languages with qf fields cover 88% of the current records vs 91.5% before the change. 

Of course, these numbers will change as we ingest new content.

PR #1132, opened 2026-04-20
Use ignore-scripts for npm install in ci

hackartisan:
also changed install to use the slightly more restricted command ci,
which will never change package.json

refs pulibrary/princeton_ansible#7007

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -125,6 +125,14 @@ defmodule DpulCollections.MixProject do
         "fixtures.setup",
         "reindex_dev"
       ],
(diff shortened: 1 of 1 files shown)


tpendragon:
@hackartisan Why is this from a fork? I think that's why CI didn't run.

hackartisan:
> @hackartisan Why is this from a fork? I think that's why CI didn't run.

oh, huh, I think maybe my gh command got messed up. thanks for catching this; I will push to the right origin and re-create

PR #1135, opened 2026-04-21
Try a lighter gray (neutral-600) on the collection page item rows

hackartisan:
the dark / light stripe on this page feels very repetitive to me, and it's a bit off-putting that the dark gray is so close in hue to the black header and footer. This is a proposal to lighten that gray a bit, to give the page a slightly less heavy look and also distinguish the content from the header / footer.

Before:

Screenshot 2026-04-21 at 10 55 19 AM

After:

Screenshot 2026-04-21 at 10 55 31 AM

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -358,7 +358,7 @@
 }
 
 .btn-arrow {
-  @apply arrow bg-accent cursor-pointer;
+  @apply arrow cursor-pointer;
 }
 
 .corner-cut {
--- a/lib/dpul_collections_web/components/browse_item.ex
+++ b/lib/dpul_collections_web/components/browse_item.ex
@@ -12,6 +12,8 @@ defmodule DpulCollectionsWeb.BrowseItem do
   attr :added?, :boolean, default: false
   attr :more_link, :string, default: nil
   attr :color, :string, default: "bg-primary-bright"
+  # can be light or dark
+  attr :arrow_theme, :string, default: "dark"
   attr :layout, :string, default: "content-area"
   attr :rest, :global
   attr :current_scope, :map, required: false, default: nil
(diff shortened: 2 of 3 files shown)


tpendragon (dismissed):

tpendragon:
Looks nice to me.

sdellis (commented):
I like the color choice, but feel that the arrow button is harder to see with the lighter grey. I don't think it's a blocker, but did we check the color contrast there?

hackartisan:
> I like the color choice, but feel that the arrow button is harder to see with the lighter grey. I don't think it's a blocker, but did we check the color contrast there?

I figured it would be caught by axe if it wasn't okay. I'll check it, though.

kelea99:
I'm not a big fan of the lighter gray, but is hard to have an opinion without seeing the entire page with the color changes and how they look as a page.   However,  I do see the benefit of the lighter gray, as the cards popping more due to the drop shadow being more noticeable. I see Shaun's point about the arrow as well.

hackartisan:
no contrast errors found when deployed on staging

sdellis:
@hackartisan I don't know why our automated Accessibility check didn't catch it, but the rust on the light grey doesn't seem to meet our criteria:

https://webaim.org/resources/contrastchecker/?fcolor=965A35&bcolor=525252

sdellis:
FYI - both current and the proposed backgrounds don't have enough contrast for AA compliance. If we use the light color from the bottom of the cards (taupe?) it will pass for both current and proposed bgs. See: 
https://webaim.org/resources/contrastchecker/?fcolor=E8E6DD&bcolor=525252

hackartisan:
PR updated with the suggested light color 

Screenshot 2026-04-23 at 10 36 56 AM

tpendragon (approved):

PR #1138, opened 2026-04-21
Removes MSS Transliterated Titles from title display in Search Results

### LINKED ISSUE #1108: Improve display of resources with transliterated titles from MMS-IDs in search results
### Summary or User Story

When a resource has two titles (because one is transliterated), they're stuffed together in the search results.

Image

### Technical Implementation

This may be because we index both values as a title, and what we should do is index the non-Latn language code as the title, and the Latn language code as `transliterated_title`. If we do so, we'll end up with a display like Ephemera, where it just displays the original script title and the transliterated titles are in a different font on item pages:

Image

We should talk about if this is the intended solution, or if (like DPUL) we should display both titles in search results but on different lines or something.

You can see the language codes here: https://figgy.princeton.edu/catalog/27fd4d29-1170-47a5-891b-f2743873bcef.jsonld (but we'll have it directly in the metadata)

`ar` means it's the original script (because it doesn't have a `-Latn`), `ar-Latn` is the transliteration.

### Acceptance Criteria

- [ ] Search results for scanned resource items with transliterated titles display better.

### First step

Update the test?

sdellis:
Fixes #1108

### DIFF
--- a/lib/dpul_collections_web/live/search_live.ex
+++ b/lib/dpul_collections_web/live/search_live.ex
@@ -512,9 +512,7 @@ defmodule DpulCollectionsWeb.SearchLive do
           id={"item-metadata-#{@item.id}"}
         >
           <div class="flex flex-wrap flex-row sm:flex-row justify-between">
-            <h2
-              dir="auto w-full flex-grow sm:w-fit"
-            >
+            <h2 dir="auto w-full flex-grow sm:w-fit">
               <.link
                 navigate={@item.url}
                 class="card-link"
--- a/lib/dpul_collections_web/live/search_live.ex
+++ b/lib/dpul_collections_web/live/search_live.ex
@@ -512,12 +512,15 @@ defmodule DpulCollectionsWeb.SearchLive do
           id={"item-metadata-#{@item.id}"}
         >
           <div class="flex flex-wrap flex-row sm:flex-row justify-between">
-            <h2 dir="auto w-full flex-grow sm:w-fit">
+            <h2
+              :for={title <- @item.title}
+              dir="auto w-full flex-grow sm:w-fit"
+            >
               <.link
                 navigate={@item.url}
                 class="card-link"
               >
-                {@item.title}
+                {title}
               </.link>
             </h2>
             <span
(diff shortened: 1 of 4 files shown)


test/dpul_collections_web/live/search_live_test.exs
```diff
@@ -798,6 +798,22 @@ defmodule DpulCollectionsWeb.SearchLiveTest do
       assert html =~ "Document-1"
     end
 
+    test "display titles and transliterated titles", %{conn: conn} do
```
tpendragon: Did this test fail before? It looks like it's just making sure that there's any title at all, I assume it had one before it was just concatenated.

lib/dpul_collections_web/live/search_live.ex
```diff
@@ -512,12 +512,15 @@ defmodule DpulCollectionsWeb.SearchLive do
           id={"item-metadata-#{@item.id}"}
         >
           <div class="flex flex-wrap flex-row sm:flex-row justify-between">
-            <h2 dir="auto w-full flex-grow sm:w-fit">
+            <h2
+              :for={title <- @item.title}
```
tpendragon: Are these changes required? Ephemera might have multiple titles, but do we have an example for this case?

tpendragon (changes requested):
Just a couple questions

lib/dpul_collections_web/live/search_live.ex
sdellis: Now that you mention it, it's not required. I was thinking that this would be a multi-valued list, but it actually just concatenates all the non-transliterated titles and will display them without looping. I will fix it up.

test/dpul_collections_web/live/search_live_test.exs
sdellis: There was no previous test to ensure that a title is displayed for a search result. However, the name of the test is misleading. I initially misunderstood the ticket to improve the display of titles and transliterations. However, the acceptance criteria was to remove transliterated titles completely. I will take that out of the test description. 👍

tpendragon (approved):

PR #1143, opened 2026-04-22
Don't query for canvas in search results / Don't parallelize

### LINKED ISSUE #1144: Filters are slow now that manuscripts are added
### Summary or User Story

When manuscripts of the islamic world got indexed all of the search results that include them got slower.

### Acceptance Criteria

- [ ] Filter performance is back to what it was.

### First step

Figure out what's slow by profiling.

tpendragon:
Parallelizing filters/items resulted in this strange disconnect between what loaded when, and performance issues were more likely related to loading all the canvas IDs for multi-hundred page manuscripts in a result set, not loading filters.

Closes #1144


hackartisan (changes requested):
What does this mean for this area of the solr code? Should it be recombined? 

https://github.com/pulibrary/dpul-collections/blob/6eb6db3720069737753d645432a0dbd4ec208e83/lib/dpul_collections/solr.ex#L124-L142

tpendragon:
> What does this mean for this area of the solr code? Should it be recombined?
> 
> https://github.com/pulibrary/dpul-collections/blob/6eb6db3720069737753d645432a0dbd4ec208e83/lib/dpul_collections/solr.ex#L124-L142

Good catch - it was definitely sending two queries, even at first mount. Reverted to pre-parallelization logic.

hackartisan (approved):

Issue #1145, opened 2026-04-22
Rendering search list after updating facets is slow


tpendragon:
### Summary or User Story

When clicking "Manuscripts of the Islamic World" from the collection facet, the view updates but it's very slow. Chrome even pauses for a bit - up to a couple seconds.

### Acceptance Criteria

- [ ] Updating the search list is performant.

### First step

Profile some things.

tpendragon:
I did `liveSocket.enableProfiling()` in the console and reloaded and I got this:

```
toString diff (update): 7.30517578125 ms
app-3d06fa3b8005a5166ec671aa976fee38.js?vsn=d:30 morphdom: 2791.05419921875 ms
app-3d06fa3b8005a5166ec671aa976fee38.js?vsn=d:30 full patch complete: 2875.109130859375 ms
```

Looking at things like https://elixirforum.com/t/liveview-performance-problems-morphdom-taking-up-to-1500ms-to-patch-updates/69408 I think the problem is probably related to updating the list of results. Components or streams might be better here.

PR #1150, opened 2026-04-22
Remove IDs from checkgroups.

tpendragon:
Significantly speeds up filters.

To figure this out I had to add a bunch of fake data locally and slowly start deleting HTML until it rendered faster.

Locally the difference is 1.2 seconds to 50ms, checking it soon on staging..

Staging (for me) is 3s to 300ms.

Work towards #1145

### DIFF
--- a/lib/dpul_collections_web/components/core_components.ex
+++ b/lib/dpul_collections_web/components/core_components.ex
@@ -684,7 +684,6 @@ defmodule DpulCollectionsWeb.CoreComponents do
       >
         <input
           type="checkbox"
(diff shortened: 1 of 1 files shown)


hackartisan (changes requested):
Wow. We didn't need these ids for anything?

tpendragon:
> Wow. We didn't need these ids for anything?

Nope, must have just been there for old versions of tests. The other input components don't have an id set by default either.

hackartisan (approved):
thanks for the thorough investigation and for sharing your process

Issue #1154, opened 2026-04-23
Don't render filter values unless they're expanded


tpendragon:
### Summary or User Story

Updating the DOM for a bunch of filters is super expensive and there's a ton of subject filters. We'd have a much faster UI if we didn't render those values to the screen unless someone actually clicks that filter.

Users rarely click filters and click subject -especially- rarely, so this will prioritize the fast experience.

### Acceptance Criteria

- [ ] Filter values are only in the DOM if that filter is expanded.
- [ ] Staging is measurably faster.
- [ ] Filters aren't less accessible.

### Technical Implementation

You can test render time on staging by going into the console, putting in `liveSocket.enableProfiling()`, and then filtering to Manuscripts of the Islamic World. The console will show data like:

```
full patch complete: 290.18505859375 ms
```

### First step

Identify where to place the logic.

PR #1156, opened 2026-04-23
Fix filter bug when submitting button

### LINKED ISSUE #1145: Rendering search list after updating facets is slow

tpendragon:
Noticed in mobile, since those inputs are gone from view now.

Fixes bug introduced in #1154 

Performance stayed the same for me.

Closes #1145

### DIFF
--- a/lib/dpul_collections_web/components/core_components.ex
+++ b/lib/dpul_collections_web/components/core_components.ex
@@ -764,6 +764,24 @@ defmodule DpulCollectionsWeb.CoreComponents do
     """
   end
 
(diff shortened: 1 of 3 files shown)


hackartisan (approved):
So if it's not expanded, it only renders hidden inputs for the current values, instead of rendering everything?

PR #1171, opened 2026-04-30
Collection Header Sizing

### LINKED ISSUE #1160: Hero image in Collection Landing Page is often cut off
### Summary or User Story

Due to the need to typographically format the text to the left of the image, the variable aspect ratio of the hero images means that it's often cut off. It would be best if it displays in its entirety.

We've also received feedback that stakeholders would like to crop images too, but displaying the full image will open a design that can accommodate more variable crops as a feature expansion.

This may require adjusting the design?

### Acceptance Criteria

- [ ] Collection landing page shows the entire image without significant cropping.

### First step

Try some CSS or designs that can accommodate more variable aspect ratios

### Example

https://digital-collections.princeton.edu/collections/lae
Image

tpendragon:
Closes #1160 

I'm not 100% on this - the extremes still end up with a bunch of dead space that feels really odd, but I do think it's better than cutting off a bunch of stuff.

It's -weirdly- hard to get an image to fill up as much space as possible, then fill in everything else beside it.

## Before

Screenshot 2026-04-30 at 4 08 22 PM

## After

Screenshot 2026-04-30 at 4 08 17 PM

## Before

Screenshot 2026-04-30 at 4 09 49 PM

## After

Screenshot 2026-04-30 at 4 10 01 PM

## Before

Screenshot 2026-04-30 at 4 10 41 PM

## After
Screenshot 2026-04-30 at 4 10 47 PM

### DIFF
--- a/lib/dpul_collections/solr.ex
+++ b/lib/dpul_collections/solr.ex
@@ -52,7 +52,8 @@ defmodule DpulCollections.Solr do
     "updated_at_dt",
     "content_warning_s",
     "geographic_origin_txt_sort",
-    "tagline_txtm"
+    "tagline_txtm",
+    "primary_thumbnail_h_w_ratio_f"
   ]
 
   def raw_query(search_state, index \\ Index.read_index()) do
--- a/lib/dpul_collections_web/live/collections_live.ex
+++ b/lib/dpul_collections_web/live/collections_live.ex
@@ -101,41 +101,40 @@ defmodule DpulCollectionsWeb.CollectionsLive do
             </h1>
           </div>
           <div class="page-y-padding relative z-30">
-            <div class="hero-container-collection flex md:flex-row-reverse w-full gap-0">
-              <!-- Right Column: Featured Items Mosaic -->
+            <div class="hero-container-collection flex flex-col md:flex-row-reverse w-full">
               <div
                 id="collection-mosaic"
-                class="flex flex-col gap-4 grow shrink-0"
+                class="flex gap-4 flex-col md:h-full md:shrink-0"
                 phx-update="ignore"
               >
-                <div class="p-2 card-darkdrop bg-white/75 flex justify-center">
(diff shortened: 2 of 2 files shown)


sdellis:
@tpendragon this is a valiant effort, and perhaps a temporary solution we need to move things forward. At the same time, I have been thinking about this in light of the conversation with @hackartisan about the preference "matted" full images and the zoomed (and randomly cropped images) for items. It's also been on my mind with the Related Collections work. 

I would like to suggest that **Items** use the full (matted) images while **Collections** (both the hero banner and cards) use intentionally cropped images. This is a subtle visual difference, but it emphasizes curation and focus that the Collections provide while providing us with some control over a consistent layout and experience for Collection pages. 

I know it adds workflow and additional feature work, but I think it will be worth it in the long run. I'm thinking that the simple Leaflet IIIF cropping tool for each "highlight" would be enough on the front-end.

tpendragon:
@sdellis Thanks for that! Not particularly valiant, just a couple hours of fiddling with CSS classes.

Are you saying that the "befores" here are better than the "afters" and we should either use cropped images or use what we have now?

I think intentional square images, or properly promotional images really, have a lot of potential. I'm happy to write the ticket, but I do think there'd be a lot more to talk about. For instance, we've recently identified that "time from curation to patron" is an important metric, and we know from DPUL that not all of our curators are interested in careful selection of the visual design of the exhibits they have today.  I think we should expect that the normal workflow is going to be adding things to a collection, sending a collection description to Kim to put in Figgy, and then that's it. If folks have to pick thumbnails, we run the risk of either collections that could have been published not happening, or Kim getting a bunch of extra work.

I might be wrong on that, I'd be interested to hear from @kelea99.

Are there particularly bad examples above? Maybe for images above a certain aspect ratio we could silently skip them or only auto-crop those.

Thanks for the discussion, see you on Tuesday!

PR #1175, opened 2026-05-01
Improve "Skip To Content"

eliotjordan:
1. Nests the "Skip To Content" in #app so tab order is correct
1. Converts "Skip To Content" to a button that focuses "#main-content. Button is more semantic.


hackartisan (changes requested):
Okay, I'm looking at the links you referenced and I'm not following the thread from those to this fix. I would like to understand and there might be more we can document about this fix. Could we pair so you can walk me through it?

eliotjordan:
No longer needed.

PR #1188, opened 2026-05-06
Ignore bandit logger errors

tpendragon:
Honeybadger for Elixir apparently has both a plug for catching errors
and can track any errors that use "Logger." Normally Honeybadger
prevents duplication by ignoring the logger domain "cowboy" which you
can see in the errors that don't have a controller and things. However,
Phoenix uses Bandit now - so replace it with "bandit" and hopefully our
duplicates go away.

Honeybadger docs here: https://github.com/honeybadger-io/honeybadger-elixir

Closes #1186

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -20,7 +20,7 @@ config :swoosh, local: false
 # Do not print debug messages in production
 config :logger, level: :info
 
-config :honeybadger, environment_name: {:system, "APP_ENV"}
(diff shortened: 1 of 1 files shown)


hackartisan (approved):
Nice find!

Here's the honeybadger plugin PR; it and the issue have more background info https://github.com/honeybadger-io/honeybadger-elixir/pull/334

PR #1189, opened 2026-05-06
[WIP] Filter preview

tpendragon:
Work towards #1185

Screenshot 2026-05-06 at 8 43 21 AM

Some notes on the implementation:

APPARENTLY `sm:` is screen based, not container based. Tailwind has "container queries", but I couldn't get them to work: https://tailwindcss.com/blog/tailwindcss-v4#container-queries

I had to switch the filter to grid to get the headers to have a consistent height. I think. There may have been a flex way, but it took a fair bit of fiddling.

@sdellis was right, a lot of duplicate ID problems. Not all solved yet.

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -278,6 +278,27 @@
   @apply text-accent hover:bg-primary-bright w-6 h-6 p-1 bg-background;
 }
 
+@utility force-mobile-items {
+  & .item > div {
+    grid-template-rows: auto !important;
+    grid-template-columns: none !important;
+  }
+
+  & .metadata {
+    grid-column: unset !important;
+    grid-column-start: unset !important;
+  }
+
+  & .small-thumbnails {
+    display: none !important;
+  }
+
+  & .search-thumbnail img {
+    height: 200px !important;
+    width: 200px !important;
+  }
+}
+
 @utility card {
   @apply drop-shadow-[0.25rem_0.25rem_0.25rem_var(--color-sage-300)];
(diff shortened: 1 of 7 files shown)


hackartisan:
Having two headings in one horizontal area feels wrong. Could we continue the columns all the way up? Or just have one heading there and then two below?

tpendragon:
> Having two headings in one horizontal area feels wrong. Could we continue the columns all the way up? Or just have one heading there and then two below?

By continue the columns all the way up, do you mean not have a preview heading?

hackartisan:
> > Having two headings in one horizontal area feels wrong. Could we continue the columns all the way up? Or just have one heading there and then two below?
> 
> By continue the columns all the way up, do you mean not have a preview heading?

By "continue the columns all the way up" I meant visually separate the top bar into two distinct areas.

can't get markdown to work here but now it's like

```
| Heading One   Heading Two|
| stuff        | subheading|
```

could we make it either like

```
| Heading One   | Heading Two|
| stuff         | subheading|
```

or like

```
| Heading for everything |
| subheading | subheading|
```

tpendragon:
@hackartisan OH, you mean when a filter is applied, because that box disappears when one isn't set. Yeah, that's a good point. 🤔

tpendragon:
Screenshot 2026-05-06 at 9 57 46 AM

@hackartisan This is what you mean?

hackartisan:
That definitely helps!

PR #1196, opened 2026-05-06
escape filter queries

### LINKED ISSUE #1164: Filters with quotes error
### Summary or User Story

[Honeybadger](https://app.honeybadger.io/projects/130995/faults/128466551/01KPQVQSNGKET608CMEH36MNQQ?q=-context.maintenance_window%3A%22true%22+occurred.after%3A%228+days+ago%22)

If you go here:

https://digital-collections.princeton.edu/i/%D1%81%D0%B5%D1%80%D0%B0%D1%84%D0%B8%D0%BC%D0%B0-%D0%B8%D0%BB%D1%8C%D0%B8%D0%BD%D0%B8%D1%87%D0%BD%D0%B0-%D0%B3%D0%BE%D0%BF%D0%BD%D0%B5%D1%80-18801966/item/ca9184a3-357d-42f1-b602-d708e278a110

and click the publisher, Solr throws an error.

Also we found that you can filter to part of a publisher and it works - that field must be stemming, and it shouldn't be for the purposes of filtering.

### Acceptance Criteria

- [ ] Filters with quotes in them works.

### First step

Replicate in a test.

hackartisan:
closes #1164

Note to reviewer: This pr could be adjusted to not add the fixture for the publisher string, and to remove the new feature test. The fixture shouldn't be needed for the solr document test, as another fixture could be used and then that value updated for the purpose of the test. The fixture and feature test were important for diagnosing and experimenting with the problem, but if string replacement is an accepted solution they could be removed without regression risk. If that's preferred, please let me know!


tpendragon (changes requested):
I wonder if this is an escape we should at query time, rather than indexing time?

tpendragon (approved):
Oh yay, it was a query time thing! I guess this is the downside of writing our own solr logic 😅

PR #1197, opened 2026-05-07
Wrap raw strings with gettext

eliotjordan:
Closes #1195 

Identifies raw strings in the website and wraps them in `gettext` for future translation.

### DIFF
--- a/lib/dpul_collections_web/components/footer_component.ex
+++ b/lib/dpul_collections_web/components/footer_component.ex
@@ -29,7 +29,7 @@ defmodule DpulCollectionsWeb.FooterComponent do
         © 2025 {gettext("The Trustees of Princeton University")}
       </div>
       <div class="university-logo flex-none w-24 sm:w-48">
-        <img src={~p"/images/university-logo.svg"} alt={gettext("Princeton University Logo")} />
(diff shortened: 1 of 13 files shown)


lib/dpul_collections_web/components/footer_component.ex
```diff
@@ -29,7 +29,7 @@ defmodule DpulCollectionsWeb.FooterComponent do
         © 2025 {gettext("The Trustees of Princeton University")}
       </div>
       <div class="university-logo flex-none w-24 sm:w-48">
-        <img src={~p"/images/university-logo.svg"} alt="Princeton University Logo" />
+        <img src={~p"/images/university-logo.svg"} alt={gettext("Princeton University Logo")} />
```
tpendragon: Can you make this `Princeton University`? Thinking about how we had to close https://github.com/pulibrary/dpul-collections/pull/1050/changes

tpendragon (dismissed):
Looks great @eliotjordan! One small thing which is kind of unrelated to the ticket, so I'd be cool with a follow up ticket and a merge here if you don't wanna deal with it.

lib/dpul_collections_web/components/footer_component.ex
eliotjordan: Done.

tpendragon (approved):

PR #1199, opened 2026-05-08
Make collection filter pills clickable

### LINKED ISSUE #1057: Limit "dismiss" click to the X potion of the filter pill
### Summary or User Story

There's an "x" on the right, but clicking anywhere on the pill dismisses the filter. Probably only the right-hand side of the pill should be clickable

formerly part of #1012

### Acceptance Criteria

- [ ] It's more clear what will happen when clicking the filter dismiss button
- [ ] The UI implemented for this issue meets accessibility requirements

### First step

ideas

### LINKED ISSUE #1158: Make collection filter pill a clickable link
### Summary or User Story

related to #1012 

As a researcher with a link to a set of search results limited to a specific collection, I need an easy way to get back to the collection page so that I can learn more about it or navigate somewhere else from there.

Sometimes you need to get from a search results page back to a collection page. We've had a stakeholder request a breadcrumb, and I have also observed this UI gap during a DLS-driven demo (the workaround used in that case was to type the url directly into the browser).

See the pill above code search results as an example of that this type of pill could look like: https://github.com/search?q=repo%3Apulibrary%2Fdpul-collections%20filter&type=code

blocked by #1057

### Acceptance Criteria

- [ ] I can easily navigate from a search result filtered to a collection to that collection itself

### First step

This section can be completed by the DLS team

Blocked by #1057

eliotjordan:
Closes #1158
Closes #1057

- **Make collection filter pills clickable**
- **Only underline the filter pill collection name**

image

### DIFF
--- a/lib/dpul_collections_web/live/search_live.ex
+++ b/lib/dpul_collections_web/live/search_live.ex
@@ -388,25 +388,53 @@ defmodule DpulCollectionsWeb.SearchLive do
 
   def filter_pill(assigns = %{filter_value: filter_value}) when is_binary(filter_value) do
     ~H"""
-    <.primary_button
+    <div
       :if={@filter_value}
-      role="button"
-      phx-value-filter-value={@filter_value}
-      phx-value-filter={@field}
-      phx-click="remove_filter"
       class={[
         @field,
-        "filter flex max-w-full gap-1 py-2 px-4 btn-primary no-underline font-semibold *:font-semibold text-sm h-full"
+        "filter inline-flex max-w-full overflow-hidden bg-primary text-dark-text font-semibold *:font-semibold text-sm h-full"
       ]}
     >
       {# These labels are defined explicitly in Solr.Constants, but have to be called here because Constants is defined at compile time.}
-      {Gettext.gettext(DpulCollectionsWeb.Gettext, @label)}
-      <span><.icon name="hero-chevron-right" class="p-1 h-4 w-4 icon" /></span>
-      <span class="filter-text truncate">
-        {@filter_value}
+      <button
+        :if={@field == "collection"}
+        type="button"
+        phx-click="navigate_to_collection"
+        phx-value-title={@filter_value}
+        class="filter-body flex items-center gap-1 py-2 pl-4 pr-2 no-underline hover:underline hover:bg-hover-accent focus-visible:underline focus-visible:bg-hover-accent cursor-pointer text-left"
+      >
+        {Gettext.gettext(DpulCollectionsWeb.Gettext, @label)}
+        <span><.icon name="hero-chevron-right" class="p-1 h-4 w-4 icon" /></span>
+        <span class="filter-text truncate">
+          {@filter_value}
+        </span>
+      </button>
+      <span
+        :if={@field != "collection"}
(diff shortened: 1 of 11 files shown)


lib/dpul_collections_web/live/search_live.ex
```diff
@@ -388,25 +388,53 @@ defmodule DpulCollectionsWeb.SearchLive do
...
     >
       {# These labels are defined explicitly in Solr.Constants, but have to be called here because Constants is defined at compile time.}
-      {Gettext.gettext(DpulCollectionsWeb.Gettext, @label)}
-      <span><.icon name="hero-chevron-right" class="p-1 h-4 w-4 icon" /></span>
-      <span class="filter-text truncate">
-        {@filter_value}
+      <button
+        :if={@field == "collection"}
+        type="button"
+        phx-click="navigate_to_collection"
```
tpendragon: Non-blocker: I wonder if this should be an href instead of a phx-click? We'd have to pre-calculate the slugs, but maybe we can do that in handle_params?

The reason would be that the semantics of a link are that the page changes, and a button stays on the same page (generally.) Also if livesocket hasn't connected yet, a button will still work.

lib/dpul_collections_web/live/search_live.ex
```diff
@@ -388,25 +388,53 @@ defmodule DpulCollectionsWeb.SearchLive do
...
-        "filter flex max-w-full gap-1 py-2 px-4 btn-primary no-underline font-semibold *:font-semibold text-sm h-full"
+        "filter inline-flex max-w-full overflow-hidden bg-primary text-dark-text font-semibold *:font-semibold text-sm h-full"
       ]}
     >
       {# These labels are defined explicitly in Solr.Constants, but have to be called here because Constants is defined at compile time.}
-      {Gettext.gettext(DpulCollectionsWeb.Gettext, @label)}
-      <span><.icon name="hero-chevron-right" class="p-1 h-4 w-4 icon" /></span>
-      <span class="filter-text truncate">
-        {@filter_value}
+      <button
```
tpendragon: Suggestion: Given the branching, should we split this out to a function component?

tpendragon (dismissed):
This works quite nicely! I have a couple suggestions - they aren't blocking, but they might make the semantics nicer.

lib/dpul_collections_web/live/search_live.ex
eliotjordan: I tried that approach, but it leads to extra Solr calls on every page load. We have to query Solr for every collection filter to get the authoritative slug. Maybe it's not so bad, but it didn't seem ideal.

TBH, this ticket seems like a workaround for not having a dedicated page / section for collections. That will resolve eventually and I didn't want to add extra blocking Solr calls at this stage. That was my thinking anyway.

Could be convinced otherwise.

tpendragon: Hmm, maybe the bigger ticket is that collection filter values should be slugs that resolve to strings in the UI.

lib/dpul_collections_web/live/search_live.ex
eliotjordan: Done.

tpendragon (dismissed):

tpendragon (approved):

PR #1218, opened 2026-05-25
Extract HydrationCacheEntry#from

tpendragon:
Small refactor that I think will make things clearer going forward. It also makes it so we can build HydrationCacheEntries, instead of attributes, which should let compile time safety tell us when attributes are wrong.


sdellis (approved):
LGTM 👍

PR #1219, opened 2026-05-25
Refactor: Dont test for noop in HydrationConsumerTest

tpendragon:
No-op batch is an implementation detail - side effects are what we're concerned with, this'll simplify these tests and make them easier to refactor later.


sdellis (approved):
Looks good! 👍

PR #1233, opened 2026-05-29
[Refactor] Playing w/ refactoring HydrationConsumer

tpendragon:
Trying out the Sandi Metz method to try refactoring HydrationConsumer, it's been interesting.

Theory is that if we back out all the classification stuff some cleaner refactors might become clear.

May not merge, just wanted to share. Commits show process.

Some refactors from here to try might be:

- [x] Extract related record handling to its own method and explicitly call it rather than have it part of the type
- [x] Extract consistent logic around "delete if exists"
- [x] Maybe extract some sort of `.process?` method?

This should be merge-able whenever, I think it's easier to read now than before and makes it clearer where opportunities to refactor are from here.

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/figgy/transformation_producer_source.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/transformation_producer_source.ex
@@ -21,7 +21,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.TransformationProducerSource do
 
     :telemetry.attach(
       "hydration-listener-#{producer_pid |> :erlang.pid_to_list()}",
-      [:broadway, :processor, :stop],
+      [:broadway, :batch_processor, :stop],
       &handle_batch_closed(&1, &2, &3, &4, cache_version),
       %{producer_pid: producer_pid}
     )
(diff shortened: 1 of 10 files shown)


hackartisan (commented):
This seems good, but I would like to better understand the goals generally and at each step. Why don't we need the batchers anymore?

The extraction of whether a record should be processed will definitely be helpful.

lib/dpul_collections/indexing_pipeline/figgy/transformation_producer_source.ex
```diff
@@ -21,7 +21,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.TransformationProducerSource do
 
     :telemetry.attach(
       "hydration-listener-#{producer_pid |> :erlang.pid_to_list()}",
-      [:broadway, :batch_processor, :stop],
+      [:broadway, :processor, :stop],
```
hackartisan: Is this change still right?

hackartisan (changes requested):
one question

lib/dpul_collections/indexing_pipeline/figgy/transformation_producer_source.ex
tpendragon: Good catch, fixed

hackartisan (approved):

PR #1236, opened 2026-05-29
Use banner image if it's been chosen

tpendragon:
Closes #1217 

Screenshot 2026-05-29 at 1 11 12 PM

Some decisions I had to make:

1. The existing implementation happens if there's no banner chosen. We can maybe change that decision?
2. If a banner image is chosen, it doesn't link to the item that banner is from. I'm sad about this, but we don't have the data anymore.


sdellis (commented):
I think this looks great. Two non-blocking comments:

1. We should apply this to the Related Collection Cards as well, but that can be a separate ticket. 
2. I am not sad about the banner link to the item going away. I think we should provide a link to the item, but we should do it via an accurate citation + link under the image.  That would be better UX and would model the preferred citations for users and suggest they use it for research/transformative uses.

tpendragon:
> 2\. I am not sad about the banner link to the item going away. I think we should provide a link to the item, but we should do it via an accurate citation + link under the image.  That would be better UX and would model the preferred citations for users and suggest they use it for research/transformative uses.

Sure, my motivation was I wanted to see the item with the cute cat, but I had no way to do so.

eliotjordan (approved):
This looks good to me. I''m in favor of merging it as is and then adding follow up tickets. To add a link to the image we would probably have to add a new field to the collection. Not sure that's great UX, but we could try.

PR #1246, opened 2026-06-03
Remove No-Op Batcher from HydrationConsumer

tpendragon:
Closes #1245

Originally the plan was to totally get rid of batchers, but that ruined performance because ack was happening a lot more often (once per message, instead of once every batch_size messages.) This still maintains batchers for performance, but doesn't require we sort them out.

Hydration time: 
02:34:23

### DIFF
--- a/config/deploy/solr/staging.json
+++ b/config/deploy/solr/staging.json
@@ -5,9 +5,9 @@
   },
   "write": [
     {
-      "cache_version": 8,
+      "cache_version": 9,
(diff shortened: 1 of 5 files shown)


hackartisan (dismissed):

hackartisan (commented):
Could you add a comment either on the PR or the issue that records why the plan changed since the issue was made?

hackartisan (approved):

PR #1251, opened 2026-06-10
Index Validation Dashboard

### LINKED ISSUE #46: Enable quick validation of per-project item counts
We often get questions about how the number of items in Figgy is different from the number of items in DPUL. These are expensive to figure out, and hard to debug. We think we can make it easier in DPUL-C by computing those differences in the dashboard.

## Success Criteria

- [ ] On the admin dashboard there's a table for every project that has the count in Figgy of EphemeraFolders vs the count in DPUL-C of indexed EphemeraFolders in that collection.
- [ ] There's also a list of folders that are in one side but not the other, with an ID, title, and a link.
- [ ] For all of the rows that DPUL-C knows it should not index, it should put them in a pre-collapsed section in that table. The UI for this is flexible, the important thing is deprioritize information that's not likely to be useful, but it can still be accessed.

## Implementation Details

We will query solr for the IDs in DPUL-C, and directly query the Figgy database for the list of Ephemera Projects and the IDs of all the folders in them.

## First Step

Write a test.

tpendragon:
Closes #46 

We made a few decisions here:

* There's really no need to see all the items that were skipped in this UI.
* Rather than try to re-check all the IDs live as part of this view, the pipeline should store information about what decisions it made for a given record and if anything went wrong. See https://github.com/pulibrary/dpul-collections/issues/1253
* The most common use case for LAE was that we wanted to filter out non-public, non-complete, and items with no member_ids - so we just did that as part of the query. We may change that when we do #1253 

Screenshot 2026-06-10 at 2 16 41 PM


lib/dpul_collections/index_validator.ex
```diff
@@ -0,0 +1,66 @@
...
+    %__MODULE__{
+      collection: collection,
+      dc_count: dc_count(collection),
+      public_complete_figgy_count: public_complete_figgy_count(collection),
+      missing_items: figgy_ids -- dc_ids,
+      extra_items: dc_ids -- figgy_ids
+    }
+  end
+
+  def dc_count(collection) do
```
hackartisan: Oh, now that we're pulling the lists of ids, we can just send the length of those instead of querying again, for both these functions

lib/dpul_collections_web/live/search_live/search_state.ex
```diff
@@ -1,4 +1,5 @@
 defmodule DpulCollectionsWeb.SearchLive.SearchState do
+  # TODO: We should pull this out and make it a struct.
```
hackartisan: Let's ticket this and remove the todo

hackartisan (dismissed):
i noticed a couple of things

eliotjordan (approved):

PR #1260, opened 2026-06-15
Cache collection thumbnails

eliotjordan:
Closes #1244

### DIFF
--- a/lib/dpul_collections/workers/cache_thumbnails.ex
+++ b/lib/dpul_collections/workers/cache_thumbnails.ex
@@ -33,7 +33,19 @@ defmodule DpulCollections.Workers.CacheThumbnails do
 
   # Don't attempt to cache deleted records, or collections
   defp cache_images(%{"deleted" => true}), do: :ok
-  defp cache_images(%{"resource_type_s" => "collection"}), do: :ok
+
+  defp cache_images(solr_document = %{"resource_type_s" => "collection"}) do
+    collection =
+      solr_document
+      |> DpulCollections.Collection.from_solr()
+
+    task = Task.async(__MODULE__, :cache_iiif_image, [collection.banner_image])
+
+    # Run task with 10 minute timeout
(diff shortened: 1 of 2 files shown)


lib/dpul_collections/workers/cache_thumbnails.ex
```diff
@@ -90,4 +102,18 @@ defmodule DpulCollections.Workers.CacheThumbnails do
...
     {:ok, %{status: 200}} = Req.request(options)
   end
+
+  def cache_iiif_collection_image(url) when is_binary(url) do
+    options =
+      [
+        url: url,
+        headers: %{"x-cache-iiif-request" => ["true"]}
+      ]
+      # Add plug option to facilitate http stubbing in tests
```
tpendragon, line 102: Suggestion: I wonder if we can write a function that takes a full URL and caches it, and use that function for both this and the above to de-dup some of this request logic? Or include it as a default in a client or something.

test/dpul_collections/workers/cache_thumbnails_test.exs
```diff
@@ -69,9 +69,46 @@ defmodule DpulCollections.Workers.CacheThumbnailsTest do
...
+            "https://example.com/iiif/2/image1/354,1295,1551,1034/750,/0/default.jpg"
+        }
+
+        # Setup ETS table to store image paths.
+        # Acts as storage for paths that are requested in different
+        # processes so they can be tested after the job has run.
+        :ets.new(:cached_paths, [:named_table, :public])
+
+        Req.Test.stub(DpulCollections.Workers.CacheThumbnails, fn conn ->
+          # Insert reqested path into ETS table
```
tpendragon:
```suggestion
          # Insert requested path into ETS table
```

tpendragon (changes requested):
Couple small suggestions, excited this seemed to slot in pretty easy.

tpendragon (approved):

PR #1267, opened 2026-06-16
Describe each fixture so we know what it is, why we have it

hackartisan:
Doing this allowed me to discover that I don't need to add a new fixture
for the feature I'm working on, maybe it will do the same down the line
for someone else


tpendragon (approved):

eliotjordan (approved):
💯

PR #1273, opened 2026-06-17
Refactors around card rows, collections, and fixtures

hackartisan:
commits should be atomic, readable
- **Move collection feature setup into each test**
- **Small fixes to md formatting for readability in editor**
- **Extract browse_li from the browse row, make it use a slot**
- **Rename some row and card components**

Moving the item cards themselves out of the row and into a slot will allow us to use the row for different kinds of cards. The cards we have (for items) are already sort of complex and they're strictly tied to the structure and data fields of an item -- e.g. the data values on an item, the creation of the image and link from the thumbnail. Trying to generalize those cards further wasn't an option.

Moving the cards out of the wrapper also allowed the wrapper to take attributes that are specific to its functionality, rather than having to take a bunch of attributes just to pass them through.

work towards #1252


tpendragon (changes requested):
The code looks fine, can you document your reasoning here a bit around converting `items` to a slot, and how it's helping that ticket?

I ask because it makes the component functionality less clear to me - I think it mostly becomes a wrapper div, but I recognize you're working with a wider context in mind.

tpendragon (approved):

PR #1277, opened 2026-06-17
1235 publisher

eliotjordan:
Closes #1235 

- **Add publisher to search result item card**
  - Moves the "Save" button to the thumbnail only in mobile view
  - Upper right side to mirror the layout in the browse cards
- **Allow the metadata elements to wrap**
- **Move genre label above title in mobile view**
- **Bold search item metadata values**

Screenshot 2026-06-17 at 4 53 37 PM
Screenshot 2026-06-17 at 4 53 15 PM

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -321,6 +321,22 @@
   }
 }
 
+/* Wrapping variant of brief-metadata. Applies the flex-wrap class and makes it so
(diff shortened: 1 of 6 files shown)


hackartisan (dismissed):
Could we try extending the matte color behind the large thumbnail down to the bottom of the row?

tpendragon (approved):

PR #1285, opened 2026-06-22
Add related collections cards

hackartisan:
closes #1252 

Note you can't see this on staging because ephemera folders can't be in more than one project, and none of the items in the one collection we are indexing on staging are ephemera folders. i.e. there are no related collections on staging.

To see this in dev, you can change the code that filters out collections from being added to item records in
/lib/dpul_collections/indexing_pipeline/figgy/resource_type_registry.ex before starting up the dev environment.
```
-  def allowed_collection?(id), do: id in @allowed_collections
+  def allowed_collection?(id), do: true
```
Then when the indexing pipeline runs it will include all the collections on the item in multiple collections. You'll be able to see this the Manuscript collection.
 

## screenshots
Screenshot 2026-06-24 at 2 57 55 PM

Screenshot 2026-06-24 at 2 58 16 PM


lib/dpul_collections_web/components/browse_item.ex
```diff
@@ -27,13 +29,11 @@ defmodule DpulCollectionsWeb.BrowseItem do
     ~H"""
     <div class={["grid-row", @color]} {@rest}>
       <div class={@layout}>
-        <div class="page-t-padding" />
-        <h2>{@title}</h2>
+        <h2 class={[@hide_title? && "sr-only", "page-t-padding"]}>{@title}</h2>
```
sdellis, line 32: It looks like Featured Highlights uses an `<h3>` here -- not sure why the Related Collections uses `<h2>`. Also, the titles appear a bit squished and don't pop to me, but this is likely another ticket and not a blocker.

lib/dpul_collections_web/live/collections_live.ex
```diff
@@ -22,20 +22,388 @@ defmodule DpulCollectionsWeb.CollectionsLive do
           assign(socket,
             page_title: collection.title,
             collection: collection,
-            mosaic_title_item: mosaic_title_item(collection)
+            banner_item: collection.banner_item
```
tpendragon, line 25: Love this.

lib/dpul_collections_web/live/collections_live.ex
```diff
@@ -22,20 +22,388 @@ defmodule DpulCollectionsWeb.CollectionsLive do
...
           )
 
         {:noreply, socket}
     end
   end
 
-  defp mosaic_title_item(%{banner_image: banner_image, banner_image_id: banner_image_id})
-       when is_binary(banner_image) and is_binary(banner_image_id) do
-    Solr.find_by_id(banner_image_id) |> Item.from_solr()
+  def render(assigns) do
```
tpendragon, line 32: I like the function components.

lib/dpul_collections_web/live/collections_live.ex
```diff
@@ -22,20 +22,388 @@ defmodule DpulCollectionsWeb.CollectionsLive do
...
+    </div>
+    """
+  end
+
+  def featured_and_related(assigns) do
+    ~H"""
+    <div>
+      <.content_separator />
+      <div class="tab-list content-area flex flex-row" role="tablist">
+        <.tab_button
```
tpendragon, line 148: We might want to split this out as a component someday.

tpendragon (dismissed):
A couple small comments, and I think we'll have to look at the design of the cards generally, but I think there's been a ton of work in here and it meets the goal of the ticket. Awesome work!

lib/dpul_collections_web/live/collections_live.ex
hackartisan: Totally, if we want to do tabs again we'll have to remember to start here.

lib/dpul_collections_web/components/browse_item.ex
hackartisan: Can you tell me where you're seeing an h3?

tpendragon (approved):

lib/dpul_collections_web/components/browse_item.ex
sdellis: @hackartisan when I inspect the Featured Highlights title, I see an `<h3>`.

PR #1290, opened 2026-06-22
Parallelize the Index Validation Dashboard

tpendragon:
This is a very rough implementation of parallelization - Flow will split the solr result into a process for every collection (up to however many cores there are), run them IO-bound, and then join them at the end.

Iteration on #46

Before this: 5.79s
After this: 2.16s


hackartisan (approved):
Cool!!

PR #1301, opened 2026-06-24
Index all published collections

tpendragon:
Closes #1298 

This works, but one thing I don't love is that I'm selecting what data we should display or not at hydration time which feels like a lot of logic, and if we change anything about those decisions we have to rehydrate.

Update: The new commit moves it to transformation, which is also a little awkward, but maybe better??


config/runtime.exs
```diff
@@ -41,7 +41,8 @@ if config_env() == :prod do
     pool_size: String.to_integer(System.get_env("POOL_SIZE") || "30"),
     # For machines with several cores, consider starting multiple pools of `pool_size`
     # pool_count: 4,
-    socket_options: maybe_ipv6
+    socket_options: maybe_ipv6,
+    timeout: 120_000
```
hackartisan, line 46: can you add a comment saying what units this is in, and maybe also why we adjusted it?

hackartisan (changes requested):
I think it makes sense. the hydration step is responsible for whether to add, update, delete a record. The transformation step is responsible for making it into a solr document.

Just one small thing inline

config/runtime.exs
tpendragon: Added, good call.

hackartisan (approved):
ty!

PR #1313, opened 2026-06-26
Collection banner images fall back to recent items.

tpendragon:
Closes #1305

I think this significantly improves the display when collections don't have highlights:
Screenshot 2026-06-26 at 3 12 09 PM


sdellis (approved):
This is great. 👍

PR #1322, opened 2026-07-01
Do not diplay Similar button on Collection page

eliotjordan:
Closes #1318


tpendragon (changes requested):
I think there are other places this appears. Can we default to false, and only set it to true in browse?

tpendragon (approved):

PR #1336, opened 2026-07-07
Update Erlang/Debian

tpendragon:
There are a bunch of security warnings from the scanner about OpenSSL
that we should be able to get past with this.

Normally https://github.com/pulibrary/dpul-collections/pull/1325 would
handle this, but CircleCI does not have Elixir 1.20 images right now.


sdellis (approved):
👍

PR #1351, opened 2026-07-15
Cache primary thumbnail and refactor

eliotjordan:
Closes #1350

- Caches primary search result thumbnail
- Refactors IIIF thumbnail definitions so the views and the cache stays in sync


hackartisan (commented):
I like this new iiif class, very nice

hackartisan (approved):

PR #1358, opened 2026-08-05
Use Consul for Elixir Clustering

tpendragon:
This PR uses https://github.com/ElixirOSS/libcluster-dynamic-srv to cluster our services, rather than using a static EPMD port. The pros for doing this:

1. Multiple instances of the elixir app can run on the same node, which should let us care less about what kind of infrastructure we're deploying onto.
2. The elixir apps have nice names
3. All the instances automatically network together.
4. We're using the SRV records of Consul to auto-manage distribution, and we can secure those ports in the future using Consul Connect.

Screenshot 2026-08-05 at 2 04 43 PM

Screenshot 2026-08-05 at 2 13 14 PM

Both the indexer and a web node running on the same VM, previously not possible.

Closes #1359


eliotjordan (approved):
💯

PR #1372, opened 2026-08-13
Display collection banner images in search results

eliotjordan:
Closes #1346 

image


tpendragon (changes requested):
I think this still squishes the banner down, and the proposal was to lock the height/width of those cards and limit the information we display to handle that.

Example of the squish:

Image

vs

Image

tpendragon (approved):

PR #1373, opened 2026-08-14
Hydrate more resources, filter them by collection during transformation

### LINKED ISSUE #1347: When a collection is changed from unpublished to published, its member resources should get indexed without anyone manually restarting any segment of the indexing pipeline
### Summary or User Story

In #1344 we made it so Figgy controls if a ScannedResource is allowed in, but we didn't make it so it gets scanned resources in that otherwise wouldn't be able to.
- we made the members allowed
- we made the hydration cache re-evaluate whether they should re-hydrate (via related items)
- But, if a thing wasn't in the hydration cache already, new resources wouldn't show up.
  - new resources means anything in the collection, because it wasn't published before.

There are a couple options to fix that - either we let all scanned resources hydrate and filter at transformation time, or we get all collection members as part of hydrating a collection and re-process them.

Note: This per-collection filtering is part of our phased roll-out, not a permanent feature

### Implementation Details

We decided that we're going to allow all ScannedResources to hydrate, and then filter (by collection) at transformation time. Filtering on visibility / state would still be in the hydration consumer (for the purposes of this ticket, at least - we can reevaluate that later).

### Acceptance Criteria

- [ ] When a collection is marked as published and it wasn't before, scanned resources that were previously filtered out should index.

### First step

Write a test.

hackartisan:
closes #1347

- **Add fixtures and test for unpublished collection**
- **Handle invalid format error in extract years**
- **Move SR collection-based filtering from hydration to transformation**

The new test is slow (about a minute); it's because this backup kicks in: 
https://github.com/pulibrary/dpul-collections/blob/eae680b91619f6924dc12df4e326778abcf55ec4/lib/dpul_collections/indexing_pipeline/database_producer.ex#L100

I think a way to make it faster would be to add an option to make that value configurable when you start a database producer. That hasn't been done in this PR. This could be blocked for that or we could make another ticket, or we could take the 1-minute hit.


hackartisan:
sending check for updates to the producer made the test run in 9s instead of 60

tpendragon (approved):

PR #1377, opened 2026-08-18
Make descriptive metadata fields searchable

### LINKED ISSUE #1107: Scanned Resource metadata fields should all be searchable
### Summary or User Story

We're now indexing a lot more fields (https://github.com/pulibrary/dpul-collections/blob/5dcdb9e3bba83eaa5f2a00db78a3175e15499813/lib/dpul_collections/indexing_pipeline/figgy/combined_figgy_resource.ex#L93-L115), but many of those fields don't return for a keyword search. 

### Acceptance Criteria

- [ ] Searching for content in any descriptive metadata field should return matching results.

### Technical Implementation

We found in #1082 that if we have too many `qf` fields it affects the performance of the application and we quickly hit the configured token maximum. As such, where normally we would just add all these fields to `qf`, we may want to copy these fields to a common field instead.

### First step

Talk about how to handle this, given #1082 and the problem with having too many `qf` fields.

eliotjordan:
Closes #1107 

Deployed and re-transformed records in staging.

**Staging** search from value in `notes_ss`:
image

**Production**:
image

### DIFF
--- a/lib/dpul_collections/indexing_pipeline/figgy/solr_document.ex
+++ b/lib/dpul_collections/indexing_pipeline/figgy/solr_document.ex
@@ -16,7 +16,7 @@ defmodule DpulCollections.IndexingPipeline.Figgy.SolrDocument do
         data: data,
         related_data: related_data
       }) do
-    build(data, related_data)
+    build(data, related_data) |> add_searchable_field()
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


lib/dpul_collections/indexing_pipeline/figgy/solr_document.ex
```diff
@@ -30,6 +30,45 @@ defmodule DpulCollections.IndexingPipeline.Figgy.SolrDocument do
...
+    case values do
+      [] -> document
+      _ -> Map.put(document, :searchable_metadata, values)
+    end
+  end
+
+  # Fields with these suffixes are used to populate searchable_metadata
+  @searchable_field_suffixes ~w(_txtm _txt_sort _ss _txt)
+
+  # Fields not to add to searchable_metadata
```
tpendragon: Interesting, is this exclusion list the reason not using a CopyField?

solr/conf/schema.xml
```diff
@@ -126,6 +126,9 @@
          because it's very expensive to index everything twice. -->
     <!-- <copyField source="*" dest="_text_"/> -->
 
+    <!-- Field to hold descriptive metadata values for search -->
+    <field name="searchable_metadata" type="text_general" indexed="true" stored="false" multiValued="true"/>
```
tpendragon: I wonder if we should use `text_en`, just because most of our cataloging is in english.

tpendragon (changes requested):
Just a couple questions.

lib/dpul_collections/indexing_pipeline/figgy/solr_document.ex
eliotjordan: Yeah. We need more control over what gets indexed in that field than we can achieve with a copyField.

solr/conf/schema.xml
eliotjordan: You mean set `type="text_en"`? That sounds reasonable.

eliotjordan: Fixed in both repos.

tpendragon (approved):

